PR #6, opened 2019-07-22
New python samples

SatjaSivcev:


.flake8
```diff
@@ -2,6 +2,7 @@
 ignore = 
     ### Disabled because of conflicts with black
     E203, # whitespace before ':'
+	F841, # local variable is assigned but not never used
```
trym-b: Write this instead:
```
### Disabled since making the samples look worse (or something like that)
    F841, # local variable is assigned but not never used
```
Notice the four spaces instead of a tab before F841

ZDF_2_CSV.py
```diff
@@ -0,0 +1,52 @@
...
+
+def _main():
+
+    app = zivid.Application()
+
+    # The Zivid3D.zdf file has to be in the same folder as this sample script.
+    FilenameZDF = "Zivid3D.zdf"
+    FilenameCSV = "Zivid3D.csv"
+
+    print("Reading ", FilenameZDF, " point cloud")
```
trym-b: if supporting 3.6 ->
`print(f"Reading {FilenameZDF} point cloud")`
else:
`print("Reading {} point cloud".format(FilenameZDF))`

ZDF_2_CSV.py
```diff
@@ -0,0 +1,52 @@
...
+    frame = zivid.Frame(FilenameZDF)
+
+    pc = frame.get_point_cloud()
+
+    # Disorganizing the point cloud
```
trym-b: Is disorganizing the right term? I do not know myself, just wondering

ZDF_2_CSV.py
```diff
@@ -0,0 +1,52 @@
...
+    pc = np.dstack(
+        [pc["x"], pc["y"], pc["z"], pc["r"], pc["g"], pc["b"], pc["contrast"]]
+    )
+    pts = pc.reshape(-1, 7)
+    # Just the points without color and contrast
+    # pc = np.dstack([pc['x'],pc['y'],pc['z']])
```
trym-b: Is it common to have commented out code in samples? Makes the samples longer and more noisy

ZDF_2_CSV.py
```diff
@@ -0,0 +1,52 @@
...
+    # pts = pc.reshape(-1,3)
+
+    # Removing nans
+    pts = pts[~np.isnan(pts[:, 0]), :]
+
+    print("Saving to a .CSV file format")
+    parts = ceil(len(pts) / 1000000)
+    np.savetxt(FilenameCSV, pts, delimiter=",", fmt="%.3f")
+    if parts > 1:
```
trym-b: if parts is not this, what should happen?

ZDF_2_CSV.py
```diff
@@ -0,0 +1,52 @@
+"""
+This example shows how to convert a Zivid point cloud from a .ZDF file format
```
trym-b: This regards all the files:
Is it obvious that this file is a sample file? Perhaps you can simplify the description to something like:
Convert ZDF point cloud to CSV format

ZDF_2_CSV_without_Zivid.py
```diff
@@ -0,0 +1,58 @@
...
+from math import ceil
+
+
+def _main():
+
+    # The Zivid3D.zdf file has to be in the same folder as this sample script.
+    FilenameZDF = "Zivid3D.zdf"
+    FilenameCSV = "Zivid3D.csv"
+
+    print("Reading ", FilenameZDF, " point cloud")
```
trym-b: f-strings or format

.flake8
nedrebo: E203 Has a comment in front of it justifying why it the test is disabled generally.

You should do the same for F841.

Right now I wonder two things:
1. Why did you not just delete the unreferenced variable
2. Why did you have to disable the warning globally and not only for the offending code.
 * Why did you globally

SatjaSivcev: I will have the comment above F841. @trym-b has mentioned it as well.
1. F.ex. one of the sample codes reads a zdf file, exports the point cloud, and then creates numpy arrays for xyz and rgb images. These variables (xyz and rgb) are not used later in the code. The point is for a user to have them available.
2. I thought that the sample code would not look nice.

capture_hdr_complete_settings.py
```diff
@@ -0,0 +1,77 @@
...
+    app = zivid.Application()
+
+    print("Connecting to the camera")
+    camera = app.connect_camera()
+
+    print("Configuring the camera settings")
+    settings_collection = [camera.settings for _ in range(3)]
+
+    settings_collection[0].iris = 10
+    settings_collection[0].exposure_time = datetime.timedelta(microseconds=10000)
```
trym-b: @nedrebo This is a use case for a custom
`__copy__`
and
`__deepcopy__`
implementation
https://stackoverflow.com/questions/1500718/how-to-override-the-copy-deepcopy-operations-for-a-python-object

capture_hdr_loop.py
```diff
@@ -0,0 +1,69 @@
...
+    settings_collection[0].filters.reflection.enabled = True
+
+    settings_collection[1].brightness = 1.8
+    settings_collection[1].gain = 1
+    settings_collection[1].filters.gaussian.enabled = True
+    settings_collection[1].filters.gaussian.sigma = 1.5
+    settings_collection[1].filters.reflection.enabled = True
+
+    settings_collection[2].brightness = 1.8
+    settings_collection[2].gain = 1.8
```
trym-b: Should this be 1?

capture_hdr_separate_frames.py
```diff
@@ -0,0 +1,57 @@
...
+    settings_collection[0].iris = 17
+    settings_collection[1].exposure_time = datetime.timedelta(microseconds=20000)
+    settings_collection[1].iris = 27
+    settings_collection[2].exposure_time = datetime.timedelta(microseconds=30000)
+    settings_collection[2].iris = 35
+
+    print("Capturing separate frames")
+    frame1 = capture(camera, settings_collection[0])
+    print(camera.settings.iris)
+    print(camera.settings.exposure_time)
```
trym-b: This, and the below values are just printed as-is. How about adding a describing text?

connect_to_serial_number_camera.py
```diff
@@ -0,0 +1,21 @@
...
+"""
+
+import zivid
+
+
+def _main():
+    app = zivid.Application()
+
+    print("Connecting to the camera")
+    camera = app.connect_camera("122026014961")
```
trym-b: perhaps make it even more explicit:
```
serial_number = "122026014961"
camera = app.connect_camera(serial_number)
```

connect_to_serial_number_camera.py
```diff
@@ -0,0 +1,21 @@
...
+
+    print(
+        "Connected to the camera with the following serial number: ",
+        camera.serial_number,
+    )
```
trym-b: f-strings or .format as commented earlier

zdf_2_csv_without_zivid.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def _main():
+
+    # The Zivid3D.zdf file has to be in the same folder as this sample script.
+    filename_zdf = "Zivid3D.zdf"
+    filename_csv = "Zivid3D.csv"
+
+    print(f"Reading {filename_zdf} point cloud")
+    data = Dataset(filename_zdf)
```
trym-b: Can you use a context manager here instead?

zdf_2_ply_without_zivid.py
```diff
@@ -1,10 +1,10 @@
 """
-This example shows how to convert a Zivid point cloud from a .ZDF file format
-to a .PLY file format.
+Convert ZDF point cloud to PLY file format without Zivid Software.
```
trym-b: without zivid software?

zdf_2_ply_without_zivid.py
```diff
@@ -46,15 +46,14 @@ def write_ply_binary(fname, pts):
...
+    filename_zdf = "Zivid3D.zdf"
+    filename_ply = "Zivid3D.ply"
 
-    # Reading a .ZDF point cloud. The "Zivid3D.zdf" file has to be in the same folder
-    # as the "ZDF2PLY" file.
-    FilenameZDF = "Zivid3D.zdf"
-    FilenamePLY = "Zivid3D.ply"
-    data = Dataset(FilenameZDF)
+    print(f"Reading {filename_zdf} point cloud")
+    data = Dataset(filename_zdf)
```
trym-b: Context manager?

zdf_2_txt_without_zivid.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def _main():
+
+    # The Zivid3D.zdf file has to be in the same folder as this sample script.
+    filename_zdf = "Zivid3D.zdf"
+    filename_txt = "Zivid3D.txt"
+
+    print(f"Reading {filename_zdf} point cloud")
+    data = Dataset(filename_zdf)
```
trym-b: context manager again?

capture_hdr_loop.py
SatjaSivcev: Yes

capture_hdr_separate_frames.py
SatjaSivcev: These were not supposed to be printed. I have deleted these lines.

connect_to_serial_number_camera.py
SatjaSivcev: Done

connect_to_serial_number_camera.py
SatjaSivcev: Done.

zdf_2_ply_without_zivid.py
SatjaSivcev: The point is that you don't need to have Zivid installed to work with ZDF files since it is based on a standard file format.

zdf_2_ply_without_zivid.py
SatjaSivcev: Done

zdf_2_txt_without_zivid.py
SatjaSivcev: Done

zdf_2_csv_without_zivid.py
SatjaSivcev: Done

trym-b:
@nedrebo Why shouldn't this pass the CI? The CI barely does anything, no samples are run. If the samples were actually run, then it should fail.

nedrebo:
> @nedrebo Why shouldn't this pass the CI?

Because you try to import a non-existing module.

PR #9, opened 2019-07-24
Zivid vtk vizualizer

SatjaSivcev:
Sample code added for point cloud visualization.


read_zdf_vis_3d.py
```diff
@@ -0,0 +1,49 @@
...
+    # Displaying the Depth Image
+    Z = xyz[:, :, 2]
+    plt.figure()
+    plt.imshow(Z, vmin=np.nanmin(Z), vmax=np.nanmax(Z), cmap="jet")
+    plt.colorbar()
+    plt.title("Depth map")
+
+
+if __name__ == "__main__":
+    # Before running this script, run '%gui qt'
```
trym-b: Have you tried running this through subprocess.Popen() ?

read_zdf_without_zivid.py
```diff
@@ -24,14 +24,16 @@ def _main():
...
     plt.imshow(rgb)
-    plt.show()
+    plt.title("RGB image")
 
     # Displaying the Depth Image
     Z = xyz[:, :, 2]
+    plt.figure()
     plt.imshow(Z, vmin=np.nanmin(Z), vmax=np.nanmax(Z), cmap="jet")
     plt.colorbar()
-    plt.show()
```
trym-b: why is show removed?

read_zdf_vis_3d.py
```diff
@@ -0,0 +1,49 @@
+"""
+Import ZDF point cloud and visualize it using vtk_visualizer (point cloud) and
+matplotlib (RGB image and Depth map).
```
trym-b: This docstring is confusing to read. Terminology is not used accurately

read_zdf_without_zivid.py
SatjaSivcev: It is not required. I think plt.imshow() already does that.

read_zdf_vis_3d.py
SatjaSivcev: Changed to 'Import ZDF point cloud and visualize it.'
The libs are probably not too relevant for the description.

read_zdf_without_zivid.py
SatjaSivcev: Actually, it was. This is why it didn't show up from the cmd. In Spyder it shows without it.

trym-b: So is this going to be reverted or not?

SatjaSivcev: we need to have plt.show() - the current version shows images.

read_zdf_vis_3d.py
SatjaSivcev: I have and it didn't work. It is not required when running via cmd prompt.

SatjaSivcev: Should we close this conversation or do you think we should try to make it work in Spyder without having to separately run that line?

read_zdf_vis_3d.py
```diff
@@ -0,0 +1,54 @@
...
+    plt.show()
+
+    # Displaying the point cloud
+    plotxyzrgb(flattened_point_cloud)
+
+
+if __name__ == "__main__":
+    # If running the script from Spyder IDE, first run '%gui qt'
+    _main()
+    input("Press Enter to close...")
```
trym-b: move this into the function

PR #11, opened 2019-07-25
Downsample

SatjaSivcev:
Sample for downsampling a point cloud. Based on Martin's Matlab code:
https://github.com/zivid/matlab-samples/blob/master/DownsamplePointCloud.m
https://github.com/zivid/matlab-samples/blob/master/downsample.m


trym-b:
You need to run black on this project

.pylintrc
```diff
@@ -0,0 +1,12 @@
+[MESSAGES CONTROL]
+disable=bad-continuation,         ######## Disabled because of conflicts with black ########
+        consider-using-enumerate, ######## Temporary disabled ########
+        import-error,
+        no-name-in-module,
+        too-many-statements,
+        unused-variable,
+
```
trym-b: I thought this file was already in master?

downsample.py
```diff
@@ -0,0 +1,163 @@
...
+        np.nansum(
+            np.transpose(np.transpose(matrix).reshape(-1, downsampling_factor)), 0
+        ).reshape(
+            int(np.shape(np.transpose(matrix))[0]),
+            int(np.shape(np.transpose(matrix))[1] / downsampling_factor),
+        )
+    )
+
+
+def downsample(xyz, rgb, contrast, downsampling_factor):
```
trym-b: perhaps add an underscore (_) for all the functions indicating that these are private functions, and not something that should be imported. Same for the other non-underscore functions.

.pylintrc
SatjaSivcev: Me too. I will do it afterward. Let's merge those first.

downsample.py
SatjaSivcev: done

SatjaSivcev:
Ready for next round

.flake8
```diff
@@ -4,6 +4,7 @@ ignore =
 	E203, # whitespace before ':'
 	### Disabled because it makes the samples look worse
 	F841, # local variable is assigned but not never used
+	W503, # line break before binary operator
```
nedrebo: Isn't this "Disabled because of conflicts with black"? Black is the one who decides formatting, not flake8.

@trym-b

PR #29, opened 2019-08-17
Add OpenCV sample

SatjaSivcev:
The sample mports a ZDF point cloud and convert it to RGB image and Depth map in OpenCV format.


SatjaSivcev:
@trym-b do you have any clue why pylint is failing?

SatjaSivcev:
Quick review, when you can please.

PR #31, opened 2019-09-12
Convert zdf

torbsorb:
Created a single script that can replace
- zdf_2_ply
- zdf_2_csv
- zdf_2_txt
It also adds an alternative to zdf_2_opencv, which is more of a "how to use ZDF with OpenCV" script.

Next up is a script to replace the conversion scripts that don't depend on zivid


convert_zdf.py
```diff
@@ -0,0 +1,168 @@
+"""Convert ZDF point cloud to the available formats.
+
+Available formats:
+    .ply - Polygon File Format
+    .csv,.txt - [X, Y, Z, r, g, b, Contrast]
+    .png,.jpg,.bmp,.tiff - 2D RGB image
```
trym-b: This is going to be out of date as soon as someone is adding a new format.

convert_zdf.py
```diff
@@ -0,0 +1,168 @@
...
+        epilog="Example: $ python -u convert_zdf.py --csv --png Zivid3D.zdf Zivid3D_1.zdf",
+    )
+
+    parser.add_argument(
+        "files", nargs="*", help="List of files to convert", default="Zivid3D.zdf"
+    )
+    parser.add_argument("--ply", action="store_true", help="Convert to PLY")
+    parser.add_argument("--csv", action="store_true", help="Convert to CSV")
+    parser.add_argument("--txt", action="store_true", help="Convert to text")
+    parser.add_argument(
```
trym-b: Are you supposed to add multiple arguments here? --txt --ply --csv ....

convert_zdf.py
```diff
@@ -0,0 +1,113 @@
...
+    )
+
+    parser.add_argument(
+        "files", nargs="*", help="List of files to convert", default="Zivid3D.zdf"
+    )
+    parser.add_argument("--ply", action="store_true", help="Convert to PLY")
+    parser.add_argument("--csv", action="store_true", help="Convert to CSV")
+    parser.add_argument("--png", action="store_true", help="Convert to PNG")
+    parser.add_argument("--txt", action="store_true", help="Convert to text")
+
```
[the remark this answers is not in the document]
trym-b: How was this resolved?

convert_zdf.py
```diff
@@ -0,0 +1,113 @@
...
+    flattened_point_cloud = point_cloud.reshape(-1, 7)
+    # Just the points without color and contrast
+    # point_cloud = np.dstack([point_cloud['x'],point_cloud['y'],point_cloud['z']])
+    # flattened_point_cloud = point_cloud.reshape(-1,3)
+
+    # Removing nans
+    return flattened_point_cloud[~np.isnan(flattened_point_cloud[:, 0]), :]
+
+
+def _convert_2_ply(frame, file_name):
```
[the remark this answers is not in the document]
trym-b: Was this not picked up by darglint?

convert_zdf.py
```diff
@@ -0,0 +1,168 @@
...
+        file_name: File name without extension
+
+    """
+    print(f"Saving the frame to {file_name}.csv")
+    np.savetxt(
+        file_name + ".csv", _flatten_point_cloud(point_cloud), delimiter=",", fmt="%.3f"
+    )
+
+
+def _convert_2_txt(point_cloud, file_name):
```
trym-b: All of these functions look very much the same, could this have been shortened?

convert_zdf.py
```diff
@@ -0,0 +1,168 @@
...
+        file_name: File name without extension
+
+    """
+    print(f"Saving the frame to {file_name}.txt")
+    np.savetxt(
+        file_name + ".txt", _flatten_point_cloud(point_cloud), delimiter=",", fmt="%.3f"
+    )
+
+
+def _convert_2_2d(point_cloud, file_name, extension):
```
trym-b: consider adding type hints

convert_zdf.py
```diff
@@ -0,0 +1,168 @@
...
+    rgb = np.dstack([point_cloud["b"], point_cloud["g"], point_cloud["r"]])
+    cv2.imwrite(f"{file_name}.{extension}", rgb)
+
+
+def _main():
+    options = _options()
+
+    app = zivid.Application()
+
+    # Loop through all provided files
```
trym-b: Unnecessary comment

convert_zdf.py
```diff
@@ -0,0 +1,168 @@
...
+    for filename_zdf in options.files:
+        if os.path.isfile(filename_zdf):
```
trym-b: Do not use os.path, use Pathlib

convert_zdf.py
```diff
@@ -0,0 +1,168 @@
...
+            if options.jpg:
+                _convert_2_2d(point_cloud, file_name, "jpg")
+
+            if options.bmp:
+                _convert_2_2d(point_cloud, file_name, "bmp")
+
+            if options.tiff:
+                _convert_2_2d(point_cloud, file_name, "tiff")
+        else:
+            print(f"{filename_zdf} not a file")
```
trym-b: raise exception

convert_zdf.py
torbsorb: Yes, I thought about refactoring as soon as I added more than one.. let me have a look. I still want the log to reflect which conversion is ongoing

convert_zdf.py
torbsorb: Will do!

convert_zdf.py
torbsorb: Sure, but what do you recommend, not list anything? Change to ".png, .jpg, etc"
Alternatively I could change the option to be a list of 2D output formats, and then simply pass them "as is" to the _convert_to_2d function.

convert_zdf.py
torbsorb: I want each to be a Boolean. It could have been a list of course, but the thought is that only a few would be requested at a time

convert_zdf.py
trym-b: Perhaps not specify it in the module documentation? For me personally I think it should already be pretty clear from the arguments in the argparse section.

convert_zdf.py
trym-b: I mean: are these mutually exclusive?

trym-b: And if they are made mutally exclusive, also make them required=True

convert_zdf.py
SatjaSivcev: I think we should have available formats written at the very top in the description area. Some customers don't care about the code and will not bother reading it. They just want to convert files and to do that they need just to know what formats are available and how to use the script.

convert_zdf.py
```diff
@@ -0,0 +1,113 @@
...
+    # Loop through all provided files
+    for filename_zdf in options.files:
+        if os.path.isfile(filename_zdf):
+            print(f"Reading {filename_zdf} point cloud")
+            frame = zivid.Frame(filename_zdf)
+            file_name = os.path.splitext(filename_zdf)[0]
+            # Getting the point cloud
+            point_cloud = frame.get_point_cloud()
+
+            # Convert based on user selection
```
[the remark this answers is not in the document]
trym-b: I think this comment is redundant. The code below is very readable.

convert_zdf.py
```diff
@@ -0,0 +1,151 @@
...
+    options = _options()
+
+    app = zivid.Application()
+
+    for filename_zdf in options.files:
+        file_path = Path(filename_zdf)
+        if file_path.is_file():
+            print(f"Reading {filename_zdf} point cloud")
+            frame = zivid.Frame(filename_zdf)
+            # Getting the point cloud
```
trym-b: I also think this is redundant

convert_zdf.py
```diff
@@ -0,0 +1,151 @@
...
+from pathlib import Path
+import numpy as np
+import cv2
+import zivid
+
+
+def _options():
+    parser = argparse.ArgumentParser(
+        description="Convert from ZDF to your preferred format",
+        epilog="Example: $ python -u convert_zdf.py --csv --png Zivid3D.zdf Zivid3D_1.zdf",
```
trym-b: Is this the common way of writing usage?

convert_zdf.py
trym-b: I see.
In that case leave it be.

convert_zdf.py
torbsorb: I don't know. I just thought it would be nice to add an example at the end. Would you rather have it before the arguments?

convert_zdf.py
torbsorb: no, they are not mutually exclusive. I did consider just adding a list of optional 2d formats, but didn't want to stray from "--format" syntax. If you believe it's better I could change it to (example) "--2d png bmp jpg"

convert_zdf.py
torbsorb: as in: should be put in each conversion function?

convert_zdf.py
```diff
@@ -0,0 +1,150 @@
...
+    group = parser.add_mutually_exclusive_group(required=True)
+    group.add_argument("--ply", action="store_true", help="Convert to PLY")
+    group.add_argument("--csv", action="store_true", help="Convert to CSV")
+    group.add_argument("--txt", action="store_true", help="Convert to text")
+    group.add_argument(
+        "--png",
+        action="store_true",
+        help="Convert to Portable Network Graphics (3D->2D)",
+    )
+    group.add_argument("--jpg", action="store_true", help="Convert to JPEG (3D->2D)")
```
trym-b: can you move this to below --txt and --tiff below this one? Just some minor formatting fixes that black does not detect

convert_zdf.py
```diff
@@ -0,0 +1,150 @@
...
+def _convert_2_ply(frame, file_name: str):
+    """Convert from frame to ply.
+
+    Args:
+        frame: A frame captured by a Zivid camera.
+        file_name: File name without extension
+
+    """
+    print(f"Saving the frame to {file_name}.ply")
+    frame.save(file_name + ".ply")
```
trym-b: I would have use an f-string here actually

convert_zdf.py
```diff
@@ -0,0 +1,150 @@
...
+    """Convert from point cloud to 2D image.
+
+    Args:
+        point_cloud: a numpy array
+        file_name: File name without extension
+
+    """
+    print(f"Saving the frame to {file_name}")
+    rgb = np.dstack([point_cloud["b"], point_cloud["g"], point_cloud["r"]])
+    cv2.imwrite(f"{file_name}", rgb)
```
trym-b: since file_name is already a string, you don't need to place it in an f-string

convert_zdf.py
```diff
@@ -0,0 +1,150 @@
...
+        elif user_options.jpg:
+            _convert_2_2d(point_cloud, file_path.stem + ".jpg")
+
+        elif user_options.bmp:
+            _convert_2_2d(point_cloud, file_path.stem + ".bmp")
+
+        elif user_options.tiff:
+            _convert_2_2d(point_cloud, file_path.stem + ".tiff")
+    else:
+        raise FileNotFoundError(f"{user_options.filename} is not a file")
```
trym-b: Looks good

convert_zdf.py
```diff
@@ -0,0 +1,150 @@
...
+    cv2.imwrite(f"{file_name}", rgb)
+
+
+def _main():
+    user_options = _options()
+
+    app = zivid.Application()
+
+    file_path = Path(user_options.filename)
+    if file_path.is_file():
```
trym-b: You don't need double nesting here. 
You can first check if the file exits (`file_path.exists()`) then do the 
`frame = zivid.Frame` 
...
`if user_options.ply:`
...

I also don't think you need to check if it is a file or not: zivid.Frame should raise an exception if it is not a zdf file.

trym-b: You should also probably check that the file exists before creating `zivid.Application`

convert_zdf.py
torbsorb: Made it mutually exclusive

convert_zdf.py
torbsorb: Just out of curiousity, why? file_name is already a string what's then the benefit of f"{file_name}.ply"? Is the latter more explicit and thus portable?

trym-b: `string1 + string2` is the old way of concatenating strings.
It usually becomes much more readable if you use a f-string:
f"{string2}{string2}"
This is a trivial example, so everything here should be more or less readable. But consider this string:
`f"useradd -m -u {uid} -g {gid} {username}"`
vs
`"useradd -m -u " + uid + " -g " + gid + " " + username"`

torbsorb: I agree with that, and in general I do use f-string

trym-b: In this case specifically:
`f"{file_name}.ply"` is clearly more obvious compared to `file_name + ".ply"`

trym-b: This is a great article about the different formatting possibilities:
https://realpython.com/python-f-strings/

torbsorb: I believe I read this, or a very similar one, a year or more ago.

convert_zdf.py
```diff
@@ -0,0 +1,150 @@
...
+    print(f"Saving the frame to {file_name}")
+    rgb = np.dstack([point_cloud["b"], point_cloud["g"], point_cloud["r"]])
+    cv2.imwrite(file_name, rgb)
+
+
+def _main():
+    user_options = _options()
+
+    file_path = Path(user_options.filename)
+    if file_path.exists():
```
trym-b: Invert this:
```
if not file_path.exists():
    raise FileNotFoundError(f"{user_options.filename} does not exist")
```
This way you don't need the additional indentation, and the error message is directly after the if-condition

convert_zdf.py
```diff
@@ -0,0 +1,150 @@
...
+    group.add_argument("--jpg", action="store_true", help="Convert to JPEG (3D->2D)")
+    group.add_argument(
+        "--png",
+        action="store_true",
+        help="Convert to Portable Network Graphics (3D->2D)",
+    )
+    group.add_argument(
+        "--bmp", action="store_true", help="Convert to Windows bitmap (3D->2D)"
+    )
+    group.add_argument("--tiff", action="store_true", help="Convert to TIFF (3D->2D)")
```
trym-b: move this one up: so that all the one that has not been broken up by black are one big block

convert_zdf.py
torbsorb: done, please verify. Wasn't sure if I did exactly what you asked for.

convert_zdf.py
torbsorb: done

convert_zdf.py
trym-b: Almost, just fix the error message, use:
`f"{user_options.filename} does not exist`

This way it is reflecting the if-condition (it is not really testing if it is a file or not, it simply tests that it exists. zivid-python is responsible for that this file is a proper zdf file, and not a dir, a png or something else)

convert_zdf.py
```diff
@@ -0,0 +1,150 @@
...
+from pathlib import Path
+import numpy as np
+import cv2
+import zivid
+
+
+def _options():
+    parser = argparse.ArgumentParser(
+        description="Convert from ZDF to your preferred format\
+            \nExample:\n\t $ python -u convert_zdf.py --ply Zivid3D.zdf",
```
trym-b: why the "-u" argument? Is this critical for this to work?

torbsorb: I believe I had it in there to force flush stdout. It was then useful for multiple conversions to show progress.

PR #33, opened 2019-09-24
Read zdf vis 3d

torbsorb:
Change 3D visualizer for the example `read_zdf_vis_3d`, from `vtk_visualizer` to `pptk`.

### DIFF
--- a/read_zdf_vis_3d.py
+++ b/read_zdf_vis_3d.py
@@ -1,11 +1,87 @@
-"""
-Import ZDF point cloud and visualize it.
-"""
+"""Import ZDF point cloud and visualize it."""
 
+import math
 import numpy as np
 import matplotlib.pyplot as plt
-from vtk_visualizer import plotxyzrgb
+import pptk
 import zivid
+
+
+def _display_rgb(rgb):
+    """Display RGB image.
+
+    Args:
+        rgb: RGB image
+
+    """
+    plt.figure()
+    plt.imshow(rgb)
+    plt.title("RGB image")
+    plt.show(block=False)
+
+
+def _display_depthmap(xyz):
+    """Create and display depthmap.
+
+    Args:
+        xyz: X, Y and Z images (point cloud co-ordinates)
+
+    """
+    plt.figure()
+    plt.imshow(
+        xyz[:, :, 2],
+        vmin=np.nanmin(xyz[:, :, 2]),
+        vmax=np.nanmax(xyz[:, :, 2]),
+        cmap="jet",
+    )
+    plt.colorbar()
+    plt.title("Depth map")
+    plt.show(block=False)
+
+
+def _get_mid_point(xyz):
+    """Calculate mid point from average of the 100 centermost points.
+
+    Args:
+        xyz: X, Y and Z images (point cloud co-ordinates)
+
+    Returns:
+        mid_point: Calculated mid point
+
+    """
+    xyz_center_cube = xyz[
+        int(xyz.shape[0] / 2 - 5) : int(xyz.shape[0] / 2 + 5),
+        int(xyz.shape[1] / 2 - 5) : int(xyz.shape[1] / 2 + 5),
+        :,
+    ]
+    return (
+        np.nanmedian(xyz_center_cube[:, :, 0]),
+        np.nanmedian(xyz_center_cube[:, :, 1]),
+        np.nanmedian(xyz_center_cube[:, :, 2]),
+    )
+
+
+def _display_pointcloud(rgb, xyz):
+    """Display point cloud.
+
+    Display the provided point cloud `xyz`, and color it with `rgb`.
+
+    We take the centermost co-ordinate as 'lookat' point. We assume that
+    camera location is at azimuth -pi/2 and elevation -pi/2 relative to
+    the 'lookat' point.
+
+    Args:
+        rgb: RGB image
+        xyz: X, Y and Z images (point cloud co-ordinates)
+
+    """
+    mid_point = _get_mid_point(xyz)
+    point_cloud_to_view = xyz
+    point_cloud_to_view[np.isnan(xyz[:, :, 2])] = 0
+    viewer = pptk.viewer(point_cloud_to_view)
+    viewer.attributes(rgb.reshape(-1, 3) / 255.0)
+    viewer.set(lookat=mid_point)
+    viewer.set(phi=-math.pi / 2, theta=-math.pi / 2, r=mid_point[2])
 
 
 def _main():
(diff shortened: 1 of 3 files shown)


read_zdf_vis_3d.py
```diff
@@ -1,38 +1,32 @@
...
-from vtk_visualizer import plotxyzrgb
 import zivid
+import pptk
 
 
-def _main():
-
-    app = zivid.Application()
+def _display_rgb(rgb):
+    """Create and display depthmap.
```
SatjaSivcev: Not depthmap but rgb image

read_zdf_vis_3d.py
```diff
@@ -1,38 +1,32 @@
...
     plt.title("RGB image")
-    plt.show()
+    plt.show(block=False)
+
+
+def _display_depthmap(xyz):
+    """Create and display depthmap.
+
+    Args:
+        xyz: X,Y and Z co-ordinates
```
SatjaSivcev: I would call this X, Y and Z Images. Co-ordinates might seem misleading - meaning three coordinates rather than 3 x 2.stg million of them.

read_zdf_vis_3d.py
```diff
@@ -1,38 +1,32 @@
...
-def _main():
-
-    app = zivid.Application()
+def _display_rgb(rgb):
+    """Create and display depthmap.
 
-    # The Zivid3D.zdf file has to be in the same folder as this sample script.
-    filename_zdf = "Zivid3D.zdf"
+    Args:
+        rgb: RGB matrix
```
SatjaSivcev: I would use RGB image for consistency with other samples. See also the comment for line 27.

read_zdf_vis_3d.py
```diff
@@ -42,14 +36,64 @@ def _main():
...
     plt.title("Depth map")
-    plt.show()
+    plt.show(block=False)
+
+
+def _get_mid_point(xyz):
+    """Calculate mid point from average of the 100 centermost points.
+
+    Args:
+        xyz: X,Y and Z co-ordinates
```
SatjaSivcev: Same as line 27.

read_zdf_vis_3d.py
```diff
@@ -42,14 +36,64 @@ def _main():
...
+    ]
+    xyz_center[xyz_center == 0] = np.nan
+    return (
+        np.nanmedian(xyz_center[:, :, 0]),
+        np.nanmedian(xyz_center[:, :, 1]),
+        np.nanmedian(xyz_center[:, :, 2]),
+    )
+
+
+def _display_pointcloud(rgb, xyz):
```
SatjaSivcev: Should we add few comments for this function? Lines 69 and 72 in particular for me were not straightforward to understand what is happening.

read_zdf_vis_3d.py
```diff
@@ -42,14 +36,64 @@ def _main():
...
+    """Calculate mid point from average of the 100 centermost points.
+
+    Args:
+        xyz: X,Y and Z co-ordinates
+
+    Returns:
+        mid_point: Calculated
+
+    """
+    xyz_center = xyz[
```
SatjaSivcev: Comment? Calculate center point or similar?

read_zdf_vis_3d.py
```diff
@@ -42,14 +36,64 @@ def _main():
...
+        int(xyz.shape[0] / 2 - 5) : int(xyz.shape[0] / 2 + 5),
+        int(xyz.shape[1] / 2 - 5) : int(xyz.shape[1] / 2 + 5),
+        :,
+    ]
+    xyz_center[xyz_center == 0] = np.nan
```
SatjaSivcev: At line 67 nans get converted to zeroes to be converted back to nans here. Is there a way to avoid this back and forth and do it only once?

read_zdf_vis_3d.py
torbsorb: I understand your comment, but I would be confused by "X, Y and Z Images". What is an "X Image"?
Maybe we could say "Set of X, Y and Z co-ordinates" or "Collection of X, Y and Z co-ordinates"

read_zdf_vis_3d.py
torbsorb: Yes, absolutely

read_zdf_vis_3d.py
SatjaSivcev: I understand your point that the terminology I suggest is not straightforward to understand for someone not familiar with it. This is the terminology we use internally and I think it makes sense to try to make customers adopt it. To me, it sounds simple and it is a nice way to split the point cloud in 3 parts, but I agree that it needs to be clarified.

How about we use both in the samples, stg like: X, Y and Z images (collections of coordinates)

SatjaSivcev: But maybe I am wrong and just used to it, I am not sure :)

torbsorb: X,Y and Z images (point cloud co-ordinates)

trym-b:
I feel like I am missing the context here, why the switch from vvtk to pptk?

SatjaSivcev:
I wanted to change it Trym. I didn't like that vtk didn't work from Spyder without running that 'qt gui' line.

torbsorb:
@trym-b, @SatjaSivcev, any more feedback? I'll rebase and squash. Then please do your final review

read_zdf_vis_3d.py
```diff
@@ -42,14 +36,76 @@ def _main():
...
+        np.nanmedian(xyz_center_cube[:, :, 2]),
+    )
+
+
+def _display_pointcloud(rgb, xyz):
+    """Display point cloud.
+
+    Display the provided point cloud `xyz`, and color it with `rgb`.
+
+    We take the centermost co-ordinate as 'lookat' point. We assume that
```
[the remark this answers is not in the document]
torbsorb: That would be good. However, I wanted to use it as a variable name, and then it can't be with a dash. In the comments I refer to the variable with the markdown notation `lookat`. Maybe I can use `look_at` instead? Or I can disregard me wanting to use the variable name in documentation.

read_zdf_vis_3d.py
```diff
@@ -42,14 +36,76 @@ def _main():
     )
     plt.colorbar()
     plt.title("Depth map")
-    plt.show()
+    plt.show(block=False)
```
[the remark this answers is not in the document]
torbsorb: Without it you have to close the figure before the script moves on. When there's only one figure then it doesn't matter. Another option is to use `plt.pause()`

read_zdf_vis_3d.py
```diff
@@ -1,38 +1,32 @@
...
-Import ZDF point cloud and visualize it.
-"""
+"""Import ZDF point cloud and visualize it."""
 
+import math
 import numpy as np
 import matplotlib.pyplot as plt
-from vtk_visualizer import plotxyzrgb
 import zivid
+import pptk
```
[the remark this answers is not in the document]
torbsorb: yes, it (will be) added to `requirements.txt`

read_zdf_vis_3d.py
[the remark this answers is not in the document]
trym-b: So this is done in order to show multiple figures at the same time?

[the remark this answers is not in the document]
torbsorb: yes

PR #34, opened 2019-09-24
Modify settings and refactor

torbsorb:
Loop over "settings [which] provide a dynamic range of some 9-10 stops, with a decent low- and highlight frame which is sufficient to grab good data on most scenes, including shiny for a medium camera operating at 0.7m-1.2m-ish"

https://zividlabs.atlassian.net/browse/CS-266

### DIFF
--- a/capture_hdr_loop.py
+++ b/capture_hdr_loop.py
@@ -1,9 +1,37 @@
-"""
-Capture HDR frames in a loop (while actively changing some HDR settings).
+"""Capture HDR frames in a loop (while actively changing HDR settings).
+
+Three sets of HDR settings are chosen. These settings provide a dynamic range of about 9-10 stops,
+with a decent lowlight and highlight frame, which is sufficient to grab good data on most scenes,
+including shiny, for a medium camera operating at around 0.7 m to 1.2 m.
+
 """
 
 import datetime
+from pathlib import Path
+import yaml
 import zivid
+
+
+def _read_settings_from_file(path: Path) -> zivid.Settings:
+    settings_from_yaml = yaml.load(path.read_text(), Loader=yaml.Loader)["Settings"]
+    settings = zivid.Settings(
+        brightness=settings_from_yaml["Brightness"],
+        exposure_time=datetime.timedelta(
+            microseconds=settings_from_yaml["ExposureTime"]
+        ),
+        gain=settings_from_yaml["Gain"],
+        iris=settings_from_yaml["Iris"],
+    )
+    filters = settings_from_yaml["Filters"]
+    settings.filters.contrast.enabled = filters["Contrast"]["Enabled"]
+    settings.filters.contrast.threshold = filters["Contrast"]["Threshold"]
+    settings.filters.gaussian.enabled = filters["Gaussian"]["Enabled"]
+    settings.filters.gaussian.sigma = filters["Gaussian"]["Sigma"]
+    settings.filters.outlier.enabled = filters["Outlier"]["Enabled"]
+    settings.filters.outlier.threshold = filters["Outlier"]["Threshold"]
+    settings.filters.reflection.enabled = filters["Reflection"]["Enabled"]
+    settings.filters.saturated.enabled = filters["Saturated"]["Enabled"]
+    return settings
 
 
 def _main():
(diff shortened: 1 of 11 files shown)


capture_hdr_loop.py
```diff
@@ -9,60 +7,47 @@
 def _main():
     app = zivid.Application()
 
-    print("Connecting to the camera")
+    print("Connect to the camera")
     camera = app.connect_camera()
+    # camera = app.create_file_camera("C:\\Program Files\\Zivid\\data\\MiscObjects.zdf")
```
SatjaSivcev: I don't think adding this here makes sense because you can just look at this emulated camera point cloud, not actually simulate taking images.

capture_hdr_loop.py
```diff
@@ -9,60 +7,47 @@
 def _main():
     app = zivid.Application()
 
-    print("Connecting to the camera")
+    print("Connect to the camera")
```
SatjaSivcev: When I started writing code samples I was using this tense and I was told that using it to print out stuff may seem as the pc is requesting things from the user, e.g. to connect to a camera or to configure the camera settings, rather than informing the user what is the pc doing atm. We need to check this with SE and figure out what to adopt as Zivid default.

capture_hdr_loop.py
```diff
@@ -9,60 +7,47 @@
...
+        datetime.timedelta(microseconds=20000),
+    )
+
+    iris_frames = (21, 17, 14)
 
     for i in range(len(iris_frames)):
-        settings_collection[0].exposure_time = exposure_time_frame_1[i]
-        settings_collection[1].exposure_time = exposure_time_frame_2[i]
-        settings_collection[2].exposure_time = exposure_time_frame_3[i]
+        settings_collection[0].exposure_time = exposure_time_frame[0] * (i + 1)
```
SatjaSivcev: I like the implementation of iterating through settings!
Curious to see the resulting images. We should check that the settings seem decent for an M camera at approx 1m for a scene with white black and colored objects.

capture_hdr_loop.py
torbsorb: Ok

capture_hdr_loop.py
torbsorb: Ok, I just used it to test out the loop (without calling camera.capture, just to check logic of the loop. Then I thought, "hey that might be useful for others" :)

capture_hdr_loop.py
torbsorb: Actually, I agree with your comment. We don't have to discuss -ing endings here.

capture_hdr_loop.py
```diff
@@ -1,68 +1,71 @@
...
+    settings.iris = iris
+    settings.gain = gain
+
+
 def _main():
     app = zivid.Application()
 
     print("Connecting to the camera")
-    camera = app.connect_camera()
+    # camera = app.connect_camera()
```
trym-b: Why is this commented out?

capture_hdr_loop.py
```diff
@@ -1,68 +1,65 @@
-"""
-Capture HDR frames in a loop (while actively changing some HDR settings).
+"""Capture HDR frames in a loop (while actively changing some HDR settings).
+
+Three sets of settings are chosen. These settings provide a dynamic range of
+about 9-10 stops, with a decent low- and highlight frame which is sufficient
```
SatjaSivcev: typo dash? 'low-' Or did you want to say low-light / lowlight?

capture_hdr_loop.py
```diff
@@ -1,68 +1,65 @@
...
+to grab good data on most scenes, including shiny for a medium camera operating
```
SatjaSivcev: maybe add a comma after 'shiny' - easier to read

capture_hdr_loop.py
```diff
@@ -1,68 +1,65 @@
...
-        datetime.timedelta(microseconds=90000),
-    ]
-    exposure_time_frame_3 = [
-        datetime.timedelta(microseconds=90000),
-        datetime.timedelta(microseconds=40000),
-        datetime.timedelta(microseconds=10000),
-    ]
-    iris_frames = [10, 20, 30]
+    # camera = app.connect_camera()
+    camera = app.create_file_camera("C:\\Program Files\\Zivid\\data\\MiscObjects.zdf")
```
SatjaSivcev: I guess you want to delete this one and uncomment line 39.

capture_hdr_loop.py
```diff
@@ -1,68 +1,65 @@
-"""
-Capture HDR frames in a loop (while actively changing some HDR settings).
+"""Capture HDR frames in a loop (while actively changing some HDR settings).
```
SatjaSivcev: We can remove 'some'?

capture_hdr_loop.py
SatjaSivcev: comma after frame? To ensure we don't mean that the hightlight frame is sufficiont to grab data on most scenes but the HDR settings

capture_hdr_loop.py
```diff
@@ -1,68 +1,65 @@
...
+
+Three sets of settings are chosen. These settings provide a dynamic range of
+about 9-10 stops, with a decent low- and highlight frame which is sufficient
+to grab good data on most scenes, including shiny for a medium camera operating
+at around 0.7m-1.2m
```
SatjaSivcev: periodat the end?

SatjaSivcev: space after number, e.g. 0.7 m to 1.2 m. ? 

P.S. Sorry, I was tortured while writing my thesis, can't help it...

capture_hdr_loop.py
```diff
@@ -1,68 +1,65 @@
...
 """
 
 import datetime
+from pathlib import Path
+import yaml
 import zivid
 
 
+def _get_settings(settings: zivid.Settings, path: Path):
+    settings_from_yaml = yaml.load(open(path, "rb").read(), Loader=yaml.Loader)["Settings"]
```
trym-b: The file opened here is actually never closed (`open(path, "rb")`)
Use `path.read_text()` instead.

capture_hdr_loop.py
```diff
@@ -1,68 +1,65 @@
...
+
 """
 
 import datetime
+from pathlib import Path
+import yaml
 import zivid
 
 
+def _get_settings(settings: zivid.Settings, path: Path):
```
trym-b: Why not simply just return a new settings instance instead of passing in an instance to be modified?

trym-b: You could also probably call this function: `_read_settings_from_file(path: Path) -> zivid.Settings:`

torbsorb: I wanted to, but wasn't sure how I would instantiate that type. I believe I figure it out, see next push.

torbsorb: I wanted to, but wasn't sure how I would instantiate that type. I believe I figure it out, see next push.

capture_hdr_loop.py
torbsorb: Hehe, I'll share blame with Theie (I copied the text from his comment in JIRA) :)

capture_hdr_loop.py
torbsorb: I shouldn't speak for Theie, but I believe the point here is that the HDR includes both a lowlight and a highlight frame? I do agree that it sounds like there's one frame that seems to capture both highlights and lowlights (which sounds counter-intuitive)...

capture_hdr_loop.py
SatjaSivcev: Does this work btw? I think maybe we should keep it but comment out. What do you think?

SatjaSivcev:
@torbsorb I think this is ready to go. Not sure if Trym's comments are resolved. Please review my changes, run it to make sure it works as you planned and check that nothing is left what you planned to modify.

SatjaSivcev:
Trym is out of office so I've added @eskaur for the final review.

capture_hdr_loop.py
```diff
@@ -1,68 +1,65 @@
...
 """
 
 import datetime
+from pathlib import Path
+import yaml
 import zivid
 
 
+def _read_settings_from_file(path: Path) -> zivid.Settings:
+    settings_from_yaml = yaml.load(path.read_text(), Loader=yaml.Loader)["Settings"]
```
trym-b: Nice use of the standard libraries

SatjaSivcev:
@eskaur can you review this so we can merge?

capture_hdr_loop.py
```diff
@@ -1,68 +1,65 @@
...
-        datetime.timedelta(microseconds=10000),
-    ]
-    iris_frames = [10, 20, 30]
-
-    for i in range(len(iris_frames)):
-        settings_collection[0].exposure_time = exposure_time_frame_1[i]
-        settings_collection[1].exposure_time = exposure_time_frame_2[i]
-        settings_collection[2].exposure_time = exposure_time_frame_3[i]
+    # Get current settings, to be modified before use
+    settings = [camera.settings for _ in range(3)]
```
eskaur: There is not need to do this. The Settings objects are overwritten in the loop below anyway. 
You may as well create an empty list within the `for hdr_index in range(3)` loop and append to it for each frame.

capture_hdr_loop.py
```diff
@@ -1,68 +1,65 @@
...
+                f" Iris: {settings[frame_index].iris}"
+                f" Exposure: {settings[frame_index].exposure_time.microseconds / 1000}ms"
+                f" Gain: {settings[frame_index].gain}"
+            )
 
-        print("Capturing an HDR frame")
-        with camera.capture(settings_collection) as hdr_frame:
-            print("Saving the HDR frame")
-            hdr_frame.save("HDR.zdf")
+        with camera.capture(settings) as hdr_frame:
```
eskaur: Consider using `zivid.hdr.capture(camera, settingslist)` instead. This is more equivalent to what we do in the main API.

capture_hdr_loop.py
torbsorb: So,
```
    # Initialize settings list
    settings = [list() for _ in range(3)]
```
?
Otherwise I'll have to create a variable to hold 3 (number of frames in the HDR).

A more radical change would be for the example to traverse the `settings/` folder and compose HDRs based on what it finds. I don't want to go there. In fact, @SatjaSivcev has proposed some other changes to this example (for C++, which we'll align with Python afterwards)

capture_hdr_loop.py
torbsorb: what's the difference?

PR #42, opened 2019-12-02
Fix get_point_cloud() due to API change

SatjaSivcev:


eskaur:
Did you run all the samples to see that they are okay?

SatjaSivcev:
Yes, that is how I was figuring out if each of them needs to be modified.

PR #46, opened 2019-12-05
add sample for hand-eye calibration on self generated dataset

chrisasc:


chrisasc:
Ready for round 1 @eskaur. Maybe we could do a live code review next week?

eskaur:
Sure @chrisasc , feel free to book a meeting-time.

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,486 @@
+"""
+Script to set up communication, generate dataset and perform hand-eye calibration.
```
eskaur: Should have more instructions for use, and perhaps list dependencies.
You forgot to update the README file, btw.

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,486 @@
...
+"""
+from pathlib import Path
+import time
+import datetime
+
+import cv2
+import numpy as np
+import zivid
+import rtde.rtde as rtde
```
eskaur: How would I install this?

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,486 @@
...
+
+    quat = np.zeros((4))
+    quat[0] = np.cos(0.5 * angle)
+    q_r = np.sqrt((1 - np.power(quat[0], 2)) / np.power(np.linalg.norm(rotvec), 2))
+    quat[1] = q_r * rotvec[0]
+    quat[2] = q_r * rotvec[1]
+    quat[3] = q_r * rotvec[2]
+    quat = quat / np.sign(quat[0])
+
+    return quat
```
eskaur: You may be re-inventing the wheel here.
Check out Scipy, with functionality such as 
https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.html

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,486 @@
...
+    Args:
+        rotvec: Rotation vector
+        angle: Angle of rotation
+
+    Returns:
+        quat: Rotation in quaternions
+    """
+
+    if np.isnan(angle):
+        angle = np.linalg.norm(rotvec)
```
eskaur: Don't use `np.nan` to represent missing optionals. Use `None`

```
def myfunc(x=None):
    if x is None:
        x = ....
```

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,486 @@
...
+
+def rotvec_to_quat(rotvec: np.array, angle: float = np.nan) -> np.array:
+    """Convert from rotation vector to quaternions
+
+    Args:
+        rotvec: Rotation vector
+        angle: Angle of rotation
+
+    Returns:
+        quat: Rotation in quaternions
```
eskaur: There's no need to name the return-values in the docstring. These names aren't visible from the callers perspective anyway.

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,486 @@
...
+    if not con.negotiate_protocol_version():
+        raise Exception(f"Protocol do not match")
+
+    if not con.send_output_setup(output_names, output_types, frequency=200):
+        raise Exception(f"Unable to configure output")
+
+    input_data = con.send_input_setup(input_names, input_types)
+
+    if not con.send_start():
+        raise Exception(f"Unable to start synchronization")
```
eskaur: We usually don't use the Exception base-class. Use something more specific, like `RuntimeError()`

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,486 @@
...
+
+
+def _set_camera_settings(cam: zivid.Camera):
+    """Set camera settings
+
+    Args:
+        cam: Zivid camera
+    """
+
+    settings = zivid.Settings()
```
eskaur: Consider doing `settings = cam.settings` to ensure that you get the defaults for the actual connected camera.

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,486 @@
...
+    settings.iris = 20
+    settings.exposure_time = datetime.timedelta(microseconds=10000)
+    settings.gain = 1
+    settings.brightness = 1.8
```
eskaur: This will fail with an Amber-camera, btw. (Max brightness of 1).

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,486 @@
...
+        output_int_register_24 = image_count:
+            Counter from robot side of number of images and poses taken.
+        output_bit_register_64 = start_capture:
+            Bool that trigger camera to capture an image
+
+        Signals to UR robot:
+        input_bit_register_64 = finish_capture:
+            Bool that trigger UR robot to move to next position.
+        input_bit_register_65 = ready_to_capture:
+            Bool that tells UR robot that the camera is ready to be used.
```
eskaur: You could consider defining this in code, perhaps as a dictionary. Would make the code where you use it more readable, as opposed to just having magic numbers.

Something like:
```
ur_registers = 
{
    "start_capture": "output_int_register_24",
    "finish_capture": "input_bit_register_64",
    ....
}

if getattr(robot_state, ur_registers["start_capture"]):
    ....
```
and `settattr()` to write.

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,486 @@
...
+            robot_state = con.receive()
+
+            print(
+                "Initial output robot_states: \n"
+                f"Image count: {robot_state.output_int_register_24} \n"
+                f"Start capture: {robot_state.output_bit_register_64}"
+            )
+
+            images_captured = 1
+            while robot_state.output_int_register_24 != -1:
```
eskaur: Stuff like this could be made much more readable (see comment above).

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,486 @@
...
+
+    Raises:
+        Exception: If computer is not able to establish comminucation with robot
+    """
+
+    conf = rtde_config.ConfigFile(Path(Path.cwd() / "robot_communication_file.xml"))
+    output_names, output_types = conf.get_recipe("out")
+    input_names, input_types = conf.get_recipe("in")
+
+    # Rtde communication is by default port 30004.
```
eskaur: This comment should probably be put where you actually set the literal "30004". (in _main())

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,486 @@
...
+
+
+def _calculate_transform_matrix(
+    robot_pose_before: np.array, robot_pose_after: np.array
+):
+    """Compute transform matrix based on poses
+
+    Args:
+        robot_pose_before: Pose of robot taken right before image was captured
+        robot_pose_after: Pose of robot right after image was captured
```
eskaur: Are these expected to be different?

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,486 @@
...
+
+            print(
+                "Initial output robot_states: \n"
+                f"Image count: {robot_state.output_int_register_24} \n"
+                f"Start capture: {robot_state.output_bit_register_64}"
+            )
+
+            images_captured = 1
+            while robot_state.output_int_register_24 != -1:
+
```
eskaur: Consider putting the contents of this loop in a separate function, for readability.

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,486 @@
...
+                    )
+                    time.sleep(0.1)
+                    _write_robot_state(
+                        con,
+                        input_data,
+                        finish_capture=False,
+                        camera_ready=ready_to_capture,
+                    )
+
+                    # Using patlib to save images
```
eskaur: Typo. Also, not sure if this comment is necessary at all :)

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,486 @@
...
+
+
+def _main():
+
+    host = "192.168.1.246"
+    port = 30004
+    con, input_data = _initialize_robot_sync(host, port)
+    con.send_start()
+
+    dataset_dir = _generate_dataset(con, input_data)
```
eskaur: Is there a good reason to save the dataset to files just to immediately re-load? Why not just pass it with variables in memory?

eskaur: It would make sense if these were two different scripts. One for generating the dataset, and one for performing the calibration computation.

hand-eye-calibration/full_hand_eye_sample.py
chrisasc: Done

hand-eye-calibration/full_hand_eye_sample.py
chrisasc: Explained in KB article, assuming the user have this installed

hand-eye-calibration/full_hand_eye_sample.py
chrisasc: Done

hand-eye-calibration/full_hand_eye_sample.py
chrisasc: Done

hand-eye-calibration/full_hand_eye_sample.py
chrisasc: Done

hand-eye-calibration/full_hand_eye_sample.py
chrisasc: Done

hand-eye-calibration/full_hand_eye_sample.py
chrisasc: Done, changed brightness to 1

hand-eye-calibration/full_hand_eye_sample.py
chrisasc: removed the comment, done.

hand-eye-calibration/full_hand_eye_sample.py
chrisasc: So small differences between one pose and the mean of two... Decided to use only one for simplicity

chrisasc: So small differences between one pose and the mean of two... Decided to use only one for simplicity

hand-eye-calibration/full_hand_eye_sample.py
chrisasc: done

hand-eye-calibration/full_hand_eye_sample.py
chrisasc: removed

hand-eye-calibration/full_hand_eye_sample.py
chrisasc: done.

hand-eye-calibration/full_hand_eye_sample.py
chrisasc: Done, created helper function for each instead.

chrisasc:
Fixed most of the changes, still need to implement user input, and check that Scipy can do the same rotations. Considering to create a helper function script because this script is too long (500+ lines).

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,539 @@
...
+    rot_matrix[1, 2] = 2 * q_y * q_z - 2 * q_x * q_w
+    rot_matrix[2, 0] = 2 * q_x * q_z - 2 * q_y * q_w
+    rot_matrix[2, 1] = 2 * q_y * q_z + 2 * q_x * q_w
+    rot_matrix[2, 2] = 1 - 2 * np.power(q_x, 2) - 2 * np.power(q_y, 2)
+
+    return rot_matrix
+
+
+def rotvec_to_quat(rotvec: np.array, angle: float = None) -> np.array:
+    """Convert from rotation vector to quaternions
```
SatjaSivcev: Use singular (quaternion)

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,539 @@
...
+    rot_matrix[1, 1] = 1 - 2 * np.power(q_x, 2) - 2 * np.power(q_z, 2)
+    rot_matrix[1, 2] = 2 * q_y * q_z - 2 * q_x * q_w
+    rot_matrix[2, 0] = 2 * q_x * q_z - 2 * q_y * q_w
+    rot_matrix[2, 1] = 2 * q_y * q_z + 2 * q_x * q_w
+    rot_matrix[2, 2] = 1 - 2 * np.power(q_x, 2) - 2 * np.power(q_y, 2)
+
+    return rot_matrix
+
+
+def rotvec_to_quat(rotvec: np.array, angle: float = None) -> np.array:
```
SatjaSivcev: It seems that this is angle-axis, not rotation vector? Or am I wrong and it can handle both?

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,539 @@
...
+import zivid
+import rtde.rtde as rtde
+import rtde.rtde_config as rtde_config
+
+
+def quat_to_rotm(quat: np.array) -> np.array:
+    """Convert from quaternion to rotation matrix
+
+    Args:
+        quat: Rotations in quaternions
```
SatjaSivcev: Use singular (rotation and quaternion)

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,539 @@
...
+        input_data.input_bit_register_65 = 0
+
+    con.send(input_data)
+
+
+def _initialize_robot_sync(host: str, port: int):
+    """Set up communication with UR robot
+
+    Args:
+        host: IP address to host
```
SatjaSivcev: I would get rid of 'to host'. It is not clear who the host is.

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,539 @@
...
+        port: Port number
+
+    Returns:
+        Connection to robot
+        Input package containing the specific input data registers
```
SatjaSivcev: It is a bit confusing when a function returns something called 'input' as it is not clear what is it input to. Maybe robot_input_data

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,539 @@
...
+The entire sample consist of two additional files:
+    - universal_robots_hand_eye_script.urp: Robot program script that moves between different poses.
+    - robot_communication_file.xml: communication set-up file.
+
+Running the sample requires that you have universal_robots_hand_eye_script.urp on your UR5e robot,
+and robot_communication_file.xml in the same repo as this sample.
+Each robot pose must be modified to your scene. This is done in universal_robots_hand_eye_script.urp on the robot.
+
+Further explanation of this sample is found in our knowledge base:
+Further explaination of this sample is found in our knowledge base:
```
SatjaSivcev: Duplicated comment lines 13 and 14

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,539 @@
+"""
+Script to generate a dataset and perform hand-eye calibration using a Universal Robot UR5e robot.
+The script communicates with the robot through TCP/IP. The host is defined by the user.
```
SatjaSivcev: The host is defined by the user. It is not clear what this means. Remove it if it is not important or further explain it, e.g. put robot IP address in brackets.

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -118,17 +82,14 @@ def _initialize_robot_sync(host: str, port: int):
         RuntimeError: If computer is not able to establish comminucation with robot
```
kontramind: Spelling error _comminucation_

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -13,69 +13,33 @@
...
+import argparse
 from pathlib import Path
 import time
 import datetime
 
 import cv2
 import numpy as np
+from scipy.spatial.transform import Rotation as R
 import zivid
 import rtde.rtde as rtde
```
kontramind: Just curious, if rtde is used -> should we as requirement (in docs) state that UR robot is required?

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -520,17 +479,23 @@ def perform_hand_eye_calibration(mode: str, data_dir: Path):
...
 def _main():
 
-    host = "192.168.1.246"
+    user_options = _options()
+
+    robot_IP_address = user_options.ip
+    # robot_IP_address = "192.168.1.246"
     port = 30004
-    con, input_data = _initialize_robot_sync(host, port)
+    con, input_data = _initialize_robot_sync(robot_IP_address, port)
```
kontramind: Will this throw an exception or return an error if the robot cannot be reached?
If so, I would suggest to handle error(s) and output an appropriate message.

chrisasc: Yes, rtde handles connection and outputs an error if it is not able to connect.

hand-eye-calibration/full_hand_eye_sample.py
chrisasc: done. using scipy isntead

hand-eye-calibration/full_hand_eye_sample.py
chrisasc: fixed

hand-eye-calibration/full_hand_eye_sample.py
chrisasc: fixed

hand-eye-calibration/full_hand_eye_sample.py
chrisasc: done

hand-eye-calibration/full_hand_eye_sample.py
chrisasc: done

hand-eye-calibration/full_hand_eye_sample.py
chrisasc: thanks, its corrected now

hand-eye-calibration/full_hand_eye_sample.py
chrisasc: Maybe, but I think you can run the script with a UR simulator a well

chrisasc:
Note: Our robots Ip adress needs to be removed before merging

hand-eye-calibration/full_hand_eye_sample.py
chrisasc: Not its a user input if they want to do generate dataset, do hand-eye or both

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,510 @@
...
+https://zivid.atlassian.net/wiki/spaces/ZividKB/pages/103481384/
+"""
+import argparse
+from pathlib import Path
+import time
+import datetime
+
+import cv2
+import numpy as np
+from scipy.spatial.transform import Rotation as R
```
eskaur: I think using the shorthand "R" here is unnecessary. It just adds another possible layer of confusion.

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,510 @@
...
+    """
+
+    frame = cam.capture()
+    robot_pose = np.array(con.receive().actual_TCP_pose)
+
+    translation = robot_pose[:3] * 1000
+    rotvec = R.from_rotvec(robot_pose[3:])
+    transform = np.eye(4)
+    transform[:3, :3] = rotvec.as_matrix()
+    transform[:3, 3] = translation.T
```
eskaur: Great job switching to scipy here. Much more neat and less error-prone :)

hand-eye-calibration/full_hand_eye_sample.py
eskaur: Seems like you fixed this :)

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,510 @@
...
+
+    if finish_capture:
+        input_data.input_bit_register_64 = 1
+    else:
+        input_data.input_bit_register_64 = 0
+
+    if camera_ready:
+        input_data.input_bit_register_65 = 1
+    else:
+        input_data.input_bit_register_65 = 0
```
eskaur: This could be simplified a lot, since `int(True)` returns 1 and `int(False)` returns 0.
The whole thing can be:
```
input_data.input_bit_register_64 = int(finish_capture)
input_data.input_bit_register_65 = int(camera_ready)
```

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,510 @@
...
+def _start_capture(robot_state):
+    """Read robot output register 64
+
+    Args:
+        robot_state: Robot state
+
+    Returns:
+        Boolean value that states if camera is ready to capture
+    """
+    return robot_state.output_bit_register_64
```
eskaur: Could this be better named as `_capture_started`? The current name makes it sound like it will do something, i.e. start a capture.

eskaur: Same for `_finish_capture`

hand-eye-calibration/full_hand_eye_sample.py
eskaur: Great :)

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,510 @@
...
+    Returns:
+        Zivid frame
+        4x4 tranformation matrix
+    """
+
+    frame = cam.capture()
+    robot_pose = np.array(con.receive().actual_TCP_pose)
+
+    translation = robot_pose[:3] * 1000
+    rotvec = R.from_rotvec(robot_pose[3:])
```
eskaur: I think `rotvec` should be renamed `rotation` here. The input is the rotation vector.

Perhaps something like this, for readability:
```
rotation_vector = robot_pose[3:]
rotation = Rotation.from_rotvec(rotation_vector)
```

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,510 @@
...
+
+    return transform, residuals
+
+
+def _main():
+
+    user_options = _options()
+
+    robot_ip_address = user_options.ip
+    # robot_ip_address = "192.168.1.246"
```
eskaur: Should probably remove this commented line before releasing the sample.

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,510 @@
...
+    """Function for taking in arguments from user
+
+    Returns:
+        Arguments from user
+
+    """
+    parser = argparse.ArgumentParser(
+        description="full_hand_eye.py [-h] [--eih] [--eth] [--ip <IP address>]",
+        formatter_class=argparse.RawTextHelpFormatter,
+    )
```
eskaur: You are not supposed to write out the brackets yourself. They are auto-generated by argparse. If you do "-h" now you'll see that it's wrong.

Do something like this?:
```
import argparse

parser = argparse.ArgumentParser(
    description='Perform Hand-Eye calibration')
mode_group = parser.add_mutually_exclusive_group(required=True)
mode_group.add_argument('--eih', "--eye-in-hand",  action='store_true',
                        help="eye-in-hand calibration")
mode_group.add_argument('--eth', "--eye-to-hand", action='store_true',
                        help="eye-to-hand calibration")
parser.add_argument("--ip", required=True,
                        help="IP address to robot")

print(parser.parse_args())
```

hand-eye-calibration/full_hand_eye_sample.py
eskaur: How would the user choose this?

hand-eye-calibration/full_hand_eye_sample.py
```diff
@@ -0,0 +1,510 @@
...
+def _generate_folder():
+    """Generate folder where dataset weill be stored
+
+    Returns:
+        Directory to where data will be saved
+    """
+
+    location_dir = Path(
+        Path.cwd() / "datasets" / datetime.datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
+    )
```
eskaur: I don't think you need to wrap this in `Path()` constructor. "Path / string" already returns a Path object.

hand-eye-calibration/full_hand_eye_sample.py
chrisasc: ready_for_capture. finish_capture is not used

hand-eye-calibration/full_hand_eye_sample.py
chrisasc: done

hand-eye-calibration/full_hand_eye_sample.py
chrisasc: done

hand-eye-calibration/full_hand_eye_sample.py
chrisasc: can simply do this to include docstring in description,

parser = argparse.ArgumentParser(description=__doc__)

eskaur: Nice

hand-eye-calibration/full_hand_eye_sample.py
chrisasc: Removed function that was not used (finish_capture and ready_to_capture)

chrisasc:
Ready for (hopefully) last round

eskaur:
I can look at this again once the tests pass :) @chrisasc

chrisasc:
Struggling to enable CI to ignore the rtde-2.3.6 scripts (as they are from UR). You can take a look at the changes on the scripts I wrote.

chrisasc:
@eskaur ready for last round now:) Squashed all commits to 1

SatjaSivcev:
@eskaur Can we merge this?

eskaur:
I see you have added `__pycache__` / `.pyc`  files to the repo. That's never a good idea, so please remove them.
This is a good reason to put a `.gitignore` file in this repo, btw :)

eskaur:
What is the purpose of the binary file `universal_robots_robot_motion_script.urp` ?

chrisasc:
universal_robots_robot_motion_script.urp is the uinversal robots program (urp) where the user sets the different poses. This script was generated using the robot control panel attached to the robot. it is easier to give customers the script than making them write it themselves. (You can load/save scripts from the robot using a USB stick).

eskaur:
> universal_robots_robot_motion_script.urp is the uinversal robots program (urp) where the user sets the different poses. This script was generated using the robot control panel attached to the robot. it is easier to give customers the script than making them write it themselves. (You can load/save scripts from the robot using a USB stick).

How can the user read/modify this script. It looks binary. Is it opened in some special UR software?

eskaur:
Otherwise the code looks great now. I can approve once my comments above are addressed, @chrisasc @SatjaSivcev

chrisasc:
Yes, it is opened in some special UR software using either a simulator or robot controller. Removed `__pycache__` 
Ready to launch this now!

eskaur:
Does RTDE still belong as a dependency in the README?  It's all bundled now, right?

SatjaSivcev:
Great work! Thanks all for contribution :)

PR #48, opened 2019-12-12
Add binary mask point cloud

SatjaSivcev:


mask_point_cloud.py
```diff
@@ -0,0 +1,129 @@
...
+
+    point_cloud = frame.get_point_cloud().to_array()
+    xyz = np.dstack([point_cloud["x"], point_cloud["y"], point_cloud["z"]])
+    rgb = np.dstack([point_cloud["r"], point_cloud["g"], point_cloud["b"]])
+
+    pixels_to_display = 300
+    print(
+        f"Generating a binary mask of central {pixels_to_display} x {pixels_to_display} pixels"
+    )
+    mask = np.zeros_like(rgb[:, :, :1])
```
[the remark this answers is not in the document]
SatjaSivcev: The intent for the mask is to be 2D. I've replaced it with: `mask = np.zeros((rgb.shape[0], rgb.shape[1], 1))`
I need it to be that shape for `xyz_masked = np.multiply(xyz, mask)` to work since xyz is of (1200,1920,3) shape.

mask_point_cloud.py
```diff
@@ -0,0 +1,129 @@
...
+    for i in range(600 - pixels_to_display, 600 + pixels_to_display):
```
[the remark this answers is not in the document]
SatjaSivcev: Done

mask_point_cloud.py
```diff
@@ -0,0 +1,129 @@
...
+
+    Args:
+        xyz: X, Y and Z images (point cloud co-ordinates)
+
+    Returns:
+        mid_point: Calculated mid point
+
+    """
+    xyz_center_cube = xyz[
+        int(xyz.shape[0] / 2 - 5) : int(xyz.shape[0] / 2 + 5),
```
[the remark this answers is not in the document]
SatjaSivcev: Done. It is a kind of offset. You get 100 central pixels by taking all pixels from position 5 left and 5 up to 5 right and 5 down.

mask_point_cloud.py
```diff
@@ -0,0 +1,129 @@
...
+
+import math
+import numpy as np
+import matplotlib.pyplot as plt
+import pptk
+import zivid
+
+
+def _display_rgb(rgb):
+    """Display RGB image.
```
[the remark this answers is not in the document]
SatjaSivcev: Do you think this is needed for the sample code? We want to make the samples as simple and easy to read as possible. I don't think it has to be production quality and super safe. Also, if we introduce checking here then it would be natural to do it elsewhere for other functions.

SatjaSivcev:
Ready for the next round.
The intent of this code sample is to demonstrate how to apply a binary mask to a point cloud. Sometimes, CAD matching or other algorithms are to be done on a region of interest (ROI) rather than the whole point cloud. One way to deal with this is to replace all pixels outside of the ROI with nan values. That is what this sample does.

mask_point_cloud.py
```diff
@@ -0,0 +1,136 @@
...
+
+    point_cloud = frame.get_point_cloud().to_array()
+    xyz = np.dstack([point_cloud["x"], point_cloud["y"], point_cloud["z"]])
+    rgb = np.dstack([point_cloud["r"], point_cloud["g"], point_cloud["b"]])
+
+    pixels_to_display = 300
+    print(
+        f"Generating a binary mask of central {pixels_to_display} x {pixels_to_display} pixels"
+    )
+    mask = np.zeros((rgb.shape[0], rgb.shape[1], 1))
```
chrisasc: You can fill your mask with nans right away by mask[...] = np.nans, then you don't have to replace the zeros afterward.

mask_point_cloud.py
```diff
@@ -0,0 +1,136 @@
...
+    height = frame.get_point_cloud().height
+    width = frame.get_point_cloud().width
+    for i in range(
+        int((height - pixels_to_display) / 2), int((height + pixels_to_display) / 2)
+    ):
+        for j in range(
+            int((width - pixels_to_display) / 2), int((width + pixels_to_display) / 2)
+        ):
+            mask[i, j] = 1
```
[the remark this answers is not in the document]
SatjaSivcev: Done.

SatjaSivcev:
> 
> 
> If the mask is a closed and aligned 2D area, then the mask could simply be:
> mask = np.ones((rgb.shape[0], rgb.shape[1]), np.bool) # First an array with all values True
> h_min = int((height - pixels_to_display) / 2)
> h_max = int((height + pixels_to_display) / 2)
> w_min = int((width - pixels_to_display) / 2)
> w_max = int((width + pixels_to_display) / 2)
> mask[h_min:h_max, w_min:w_max] = 0 # Then we set what we want to keep to False
> and masking would be:
> xyz_masked = xyz.copy()
> xyz_masked[mask] = np.nan



> 
> 
> If the mask is a closed and aligned 2D area, then the mask could simply be:
> mask = np.ones((rgb.shape[0], rgb.shape[1]), np.bool) # First an array with all values True
> h_min = int((height - pixels_to_display) / 2)
> h_max = int((height + pixels_to_display) / 2)
> w_min = int((width - pixels_to_display) / 2)
> w_max = int((width + pixels_to_display) / 2)
> mask[h_min:h_max, w_min:w_max] = 0 # Then we set what we want to keep to False
> and masking would be:
> xyz_masked = xyz.copy()
> xyz_masked[mask] = np.nan

I've modified the code to use this approach. However, I've inverted the logic. It is more common (and OpenCV is using such logic) that '1' represents the pixels that stay and '0' the ones that are masked away, so I want to avoid any confusion.

mask_point_cloud.py
chrisasc: btw, do you need to use NaNs at all as they are set to 0 again when you display the point cloud

SatjaSivcev: I've removed this line in favor of `mask = np.zeros((rgb.shape[0], rgb.shape[1]), np.bool)`
Can I still avoid the line? `xyz_masked[mask == 0] = np.nan` I don't see how.

SatjaSivcev:
Ready for the next round

mask_point_cloud.py
chrisasc: I was a bit quick there, you are right, you cannot avoid `xyz_masked[mask == 0] = np.nan`, sorry!

SatjaSivcev:
Reminder for next round of review

SatjaSivcev:
@kontramind - a reminder for the next round of review

source/applications/basic/visualization/read_zdf_vis_3d.py
```diff
@@ -75,6 +75,8 @@ def _display_pointcloud(rgb, xyz):
         rgb: RGB image
         xyz: X, Y and Z images (point cloud co-ordinates)
 
+    Returns None
+
```
torbsorb: Should be separate commit. I assume this is because CI fails? Did you rebase first? Is so, then this should be the first commit.

SatjaSivcev: I've split the commits and placed this one on the first place.

SatjaSivcev:
I've added the sample to the README. One last review @torbsorb

SatjaSivcev:
Can we merge this? It's been idle for too long.

SatjaSivcev:
Rebased, need approval.

mask_point_cloud.py
```diff
@@ -0,0 +1,129 @@
...
+    xyz = np.dstack([point_cloud["x"], point_cloud["y"], point_cloud["z"]])
+    rgb = np.dstack([point_cloud["r"], point_cloud["g"], point_cloud["b"]])
+
+    pixels_to_display = 300
+    print(
+        f"Generating a binary mask of central {pixels_to_display} x {pixels_to_display} pixels"
+    )
+    mask = np.zeros_like(rgb[:, :, :1])
+    for i in range(600 - pixels_to_display, 600 + pixels_to_display):
+        for j in range(960 - pixels_to_display, 960 + pixels_to_display):
```
trym-b, line 107: https://github.com/zivid/python-samples/pull/48/files#r357496991
Perhaps to the same here with 960, give them a name. Same for 600 above.

mask_point_cloud.py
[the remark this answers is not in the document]
trym-b: These should probably have a name, since these numbers are now magic numbers. Giving them a label helps the reader of the code where these numbers came from, and also reduces code duplication

SatjaSivcev:
I already made those changes @trym-b, maybe I forgot to push or something... ready for review.

PR #38, opened 2019-10-26
Add utilize eye in hand calibration sample

SatjaSivcev:

PR #49, opened 2020-01-03
Add example for conversion of rotation formats

torbsorb:
This PR depends on #38. 

The example is the python equivalent of [PoseConversions.cpp](https://github.com/zivid/cpp-extra-samples/blob/master/HandEyeCalibration/PoseConversions/PoseConversions.cpp). 

Some of the code will look a bit strange, such as:
`robot_representations.axis_angle = AxisAngle(rotation.as_rotvec())`
where the conversion from Rotation Matrix to Axis-Angle goes via rotation vector.
and maybe even worse:
`quaternion = R.from_rotvec(axis_angle.as_rotvec()).as_quat()`
where the goal was to go from Axis-Angle to Quaternion, but I ended up going via rotation vector and scipy's internal rotation representation.

The reason is to avoid re-implementing all conversions, and rather reuse an available python library. Currently scipy.spatial.transform.Rotations is the best. I've investigated the following:
- **Open3D** - only provide transformations **to** Rotation Matrix
- **transforms3d** - has everything, but throws `ValueError: no unit eigenvector corresponding to eigenvalue 1` on `t3d.axangles.aff2axangle(transformation_matrix)`. It throws the same on `tmp = t3d.axangles.mat2axangle(transformation_matrix[:-1, :-1])`. This error is not thrown by `scipy.spatial.transform.Rotations.from_matrix()`

### DIFF
--- a/pose_conversions.py
+++ b/pose_conversions.py
@@ -1,0 +1,283 @@
+"""Example to show conversions to/from Transformation Matrix.
+
+Zivid primarily operate with a (4x4) Transformation Matrix (Rotation Matrix + Translation Vector).
+This example shows how to use Eigen to convert to and from:
+  AxisAngle, Rotation Vector, Roll-Pitch-Yaw, Quaternion
+
+ It provides convenience functions that can be reused in applicable applications.
+"""
+
+import enum
+from pathlib import Path
+from dataclasses import dataclass, field
+import numpy as np
+import cv2
+from scipy.spatial.transform import Rotation as R
+
+
+def _main():
+    np.set_printoptions(precision=4, suppress=True)
+    print_header(f"This example shows conversions to/from Transformation Matrix")
+
+    transformation_matrix = get_transformation_matrix_from_yaml("robotTransform.yaml")
+    print(f"Transformation Matrix:\n{transformation_matrix}")
+
+    # Extract Rotation Matrix and Translation Vector from Transformation Matrix
+    print(f"Rotation Matrix:\n{transformation_matrix[:3,:3]}")
+    print(f"Translation Vector:\n{transformation_matrix[:-1, -1]}")
+
+    ###
+    # Convert from Zivid to Robot (Transformation Matrix --> any format)
+    ###
+    print_header("Convert from Zivid (Rotation Matrix) to Robot")
+    axis_angle = rotation_matrix_to_axis_angle(transformation_matrix[:3, :3])
+    print(f"AxisAngle:\n{axis_angle.axis}, {axis_angle.angle:.4f}")
+    rotation_vector = rotation_matrix_to_rotation_vector(transformation_matrix[:3, :3])
+    print(f"Rotation Vector:\n{rotation_vector}")
+    quaternion = rotation_matrix_to_quaternion(transformation_matrix[:3, :3])
+    print(f"Quaternion:\n{quaternion}")
+    rpy_list = rotation_matrix_to_roll_pitch_yaw(transformation_matrix[:3, :3])
+
+    ###
+    # Convert from Robot to Zivid (any format --> Rotation Matrix (part of Transformation Matrix))
+    ###
+    print_header("Convert from Robot to Zivid (Rotation Matrix)")
+    rotation_matrix = axis_angle_to_rotation_matrix(axis_angle)
+    print(f"Rotation Matrix from Axis Angle:\n{rotation_matrix}")
+    rotation_matrix = rotation_vector_to_rotation_matrix(rotation_vector)
+    print(f"Rotation Matrix from Rotation Vector:\n{rotation_matrix}")
+    rotation_matrix = quaternion_to_rotation_matrix(quaternion)
+    print(f"Rotation Matrix from Quaternion:\n{rotation_matrix}")
+    roll_pitch_yaw_to_rotation_matrix(rpy_list)
+
+    # Replace rotation matrix in transformation matrix
+    transformation_matrix[:3, :3] = rotation_matrix
+    # Save transformation matrix which has passed through quaternion representation
+    save_transformation_matrix_to_yaml(transformation_matrix, "robotTransformOut.yaml")
+
+
+class RotationConvention(enum.Enum):
+    """Convenience enum class to list rotation conventions for Roll Pitch Yaw."""
+
+    ZYX_Intrinsic = "ZYX"
+    XYZ_Extrinsic = "xyz"
+    XYZ_Intrinsic = "XYZ"
+    ZYX_Extrinsic = "zyx"
+
+
+class AxisAngle:
+    """Convenience class to access rotation axis and angle."""
+
+    def __init__(self, axis=np.array([0, 0, 1]), angle=None):
+        """Initialize class and its variables.
+
+        Can be initialized with a unit vector and an angle, or only a rotation vector.
+
+        Args:
+            axis: rotation axis
+            angle: rotation angle
+
+        Raises:
+            ValueError: if angle vector is provided, but vector is not a unit vector
+
+        """
+        self.angle = angle
+        self.axis = axis
+        if angle is None:
+            self.angle = np.linalg.norm(axis)
+            self.axis = axis / self.angle
+        elif np.linalg.norm(axis) != 0:
+            raise ValueError(f"Angle provided, but vector is not unit vector")
+
+    def as_rotvec(self):
+        """Return rotation vector from axis angle.
+
+        Returns:
+            rotation vector
+
+        """
+        return self.axis * self.angle
+
+    def as_quaternion(self):
+        """Return quaternion from axis angle.
+
+        Returns:
+            quaternion
+
+        """
+        return R.from_rotvec(self.as_rotvec()).as_quat()
+
+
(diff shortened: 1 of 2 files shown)


torbsorb:
@trym-b and @SatjaSivcev , it's clean and ready for review now, sorry about the spam.

trym-b:
I would like this file to be similar to this sample you have already created:
https://github.com/zivid/python-samples/blob/master/convert_zdf.py

Each conversion is done in a separate function that does exactly what it supposed to do, no more, no less. If the sample could look something like this, I would be happy.

torbsorb:
@SatjaSivcev, 
@trym-b questioned why we do all conversion between non-transformation matrix formats? He's arguing that we should only show from Zivid to Robot and Robot to Zivid, and not all conversions between Robot formats. I couldn't argue well against. If we did that then the code could be cleaned up a little bit.

trym-b:
Have you come to any conclusions yet? Is the current state of this PR what you would like to have in the repo? @SatjaSivcev  @torbsorb ?
If you decide this is the way to go I won't stand in your way, but just note that some of the samples are starting to diverge in structure/form.

chrisasc:
@trym-b Can we merge this as well? It was dependent on #38 which is now merged

SatjaSivcev:
> 
> 
> @SatjaSivcev,
> @trym-b questioned why we do all conversion between non-transformation matrix formats? He's arguing that we should only show from Zivid to Robot and Robot to Zivid, and not all conversions between Robot formats. I couldn't argue well against. If we did that then the code could be cleaned up a little bit.

Isn't it inevitable to do this? If you want to go from e.g. transform to rotation vector, you have to first convert to angle axis and then to rotation vector. Is the alternative to hide the in-between conversions? I don't think I understand what @trym-b meant.

SatjaSivcev:
@torbsorb, @trym-b explained what he means in person. I am convinced. I think it would be more clear if we do it that way. Let's sit down and agree on what goes into main and how we make the functions.

torbsorb:
@trym-b 
Made changes according to your comments

pose_conversions.py
```diff
@@ -0,0 +1,262 @@
...
+    ###
+    # Convert from Robot to Zivid (any format --> Rotation Matrix (part of Transformation Matrix))
+    ###
+    print_header("Convert from Robot to Zivid (Rotation Matrix)")
+    rotation_matrix = axis_angle_to_rotation_matrix(axis_angle)
+    print(f"Rotation Matrix from Axis Angle:\n{rotation_matrix}")
+    rotation_matrix = rotation_vector_to_rotation_matrix(rotation_vector)
+    print(f"Rotation Matrix from Rotation Vector:\n{rotation_matrix}")
+    for rotation in rpy_list:
+        rotation_matrix_from_roll_pitch_yaw = R.from_euler(
```
SatjaSivcev: In 55 and 57 you use rotation_matrix. Here you use rotation_matrix_from_roll_pitch_yaw. Either reuse the same variable rotation_matrix or make them all unique e.g. rotation_matrix_from_...

pose_conversions.py
```diff
@@ -0,0 +1,262 @@
...
+    ###
+    print_header("Convert from Zivid to Robot")
+    axis_angle = zivid_to_axis_angle(transformation_matrix)
+    print(f"AxisAngle:\n{axis_angle.angle}, {axis_angle.axis}")
+
+    rotation_vector = zivid_to_rotation_vector(transformation_matrix)
+    print(f"Rotation Vector:\n{rotation_vector}")
+
+    quaternion = zivid_to_quaternion(transformation_matrix)
+    print(f"Qaternion:\n{quaternion}")
```
trym-b: I like this part.

pose_conversions.py
```diff
@@ -0,0 +1,262 @@
...
+
+    rpy_list = list()
+    rotation_matrix = transformation_matrix[:3, :3]
+    rotation = R.from_matrix(rotation_matrix)
+    for convention in RotationConvention:
+        roll_pitch_yaw = rotation.as_euler(convention.value)
+        print(f"Roll-Pitch-Yaw angles ({convention.name}):")
+        print(f"{roll_pitch_yaw}")
+        rpy_list.append({"convention": convention, "roll_pitch_yaw": roll_pitch_yaw})
```
trym-b: Could this be moved to a function?

pose_conversions.py
```diff
@@ -0,0 +1,262 @@
...
+
+    ###
+    # Convert from Robot to Zivid (any format --> Rotation Matrix (part of Transformation Matrix))
+    ###
+    print_header("Convert from Robot to Zivid (Rotation Matrix)")
+    rotation_matrix = axis_angle_to_rotation_matrix(axis_angle)
+    print(f"Rotation Matrix from Axis Angle:\n{rotation_matrix}")
+    rotation_matrix = rotation_vector_to_rotation_matrix(rotation_vector)
+    print(f"Rotation Matrix from Rotation Vector:\n{rotation_matrix}")
```
trym-b: Nice

pose_conversions.py
```diff
@@ -0,0 +1,262 @@
...
+    for rotation in rpy_list:
+        rotation_matrix_from_roll_pitch_yaw = R.from_euler(
+            rotation["convention"].value, rotation["roll_pitch_yaw"]
```
trym-b: Could this also be moved to a function?

pose_conversions.py
```diff
@@ -0,0 +1,262 @@
...
+        ).as_matrix()
+        print(
+            f"Rotation Matrix from Roll-Pitch-Yaw angles ({rotation['convention'].name})"
+        )
+        print(f" to :\n{rotation_matrix_from_roll_pitch_yaw}")
+
+    # Replace rotation matrix in transformation matrix
+    transformation_matrix[:3, :3] = rotation_matrix
+    # Save transformation matrix which has passed through quaternion representation
+    save_transformation_matrix_to_yaml(transformation_matrix, "robotTransformOut.yaml")
```
trym-b: Could the save to yaml be in a different function?

pose_conversions.py
```diff
@@ -0,0 +1,262 @@
...
+        path: path to the YAML file
+
+    Returns:
+        4x4 Transformation Matrix
+
+    """
+    file_storage_in = cv2.FileStorage(path, cv2.FILE_STORAGE_READ)
+    transformation_matrix = file_storage_in.getNode("PoseState").mat()
+    file_storage_in.release()
+    return transformation_matrix
```
trym-b: Looks good.

pose_conversions.py
```diff
@@ -0,0 +1,262 @@
...
+    """Print decorated header.
+
+    Args:
+        txt: Text to be printed in header
+
+    """
+    print()
+    print(f"{'*' * 70}")
+    print(f"* {txt} {' ' * (70 - len(txt) - 4)}*")
+    print(f"{'*' * 70}")
```
trym-b: Not sure what the purpose of the this function is. Can you show an example print here in this PR?

pose_conversions.py
```diff
@@ -0,0 +1,262 @@
...
+
+    rotation_vector = zivid_to_rotation_vector(transformation_matrix)
+    print(f"Rotation Vector:\n{rotation_vector}")
+
+    quaternion = zivid_to_quaternion(transformation_matrix)
+    print(f"Qaternion:\n{quaternion}")
+
+    rpy_list = list()
+    rotation_matrix = transformation_matrix[:3, :3]
+    rotation = R.from_matrix(rotation_matrix)
```
SatjaSivcev: Perhaps replace 43 and 44 with rotation = R.from_matrix(transformation_matrix[:3, :3])?

pose_conversions.py
```diff
@@ -0,0 +1,262 @@
...
+class Representations:
+    """Class to hold various transformation representations."""
+
+    axis_angle: AxisAngle() = AxisAngle()
+    rotation_vector: np.array = np.zeros(3)
+    quaternion: np.array = np.zeros(4)
+    rotations: list = field(default_factory=list)
+
+
+def zivid_to_axis_angle(transformation_matrix):
```
SatjaSivcev: I would name these type of functions as transformation_matrix_to_something as it is not consistend that in one direction you have zivid_to_something but not in the other direction i.e. something_to_zivid

pose_conversions.py
torbsorb: Is variable re-usage in python ok @trym-b ? I know Anders was against it in C++

torbsorb: Solved when extracted as a method

pose_conversions.py
torbsorb:
```
print_header("Convert from Zivid to Robot")
```

yields:

```
**********************************************************************
* Convert from Zivid to Robot                                        *
**********************************************************************
```

pose_conversions.py
torbsorb: solved through refactor

SatjaSivcev:
Shall we merge so? @torbsorb

torbsorb:
@SatjaSivcev, rebased and fixed two typos. Please approve and I'll merge.

PR #55, opened 2020-03-13
Add two Capture tutorials

SatjaSivcev:
Two tutorials for Capture is added. One thorough and one quick.


SatjaSivcev:
I did a bit more than just clone the C++ version. The differences:
Added Save 2D section
Replaced and added more (go to source)
Modified some of the code in the tutorial relative to the code in the samples (code in samples is to be modified to be the same as the code in the tutorial in different PRs)

SatjaSivcev:
@torbsorb can you approva again so I can merge?

PR #60, opened 2020-03-25
Gamma correction sample

torbsorb:
Captured point clouds generally tend to have very dark colors. In order to better visualize both the point cloud and the 2D image, we can adjust the gamma setting. This sample shows how this is done.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,9 @@ There are two main categories of samples: **camera** and **applications**. The s
     - [**create_depth_map**][create_depth_map-url] - Import a ZDF point cloud and convert it to OpenCV format, then extract and visualize depth map.
       - **Dependencies:**
         - [OpenCV](https://opencv.org/) version 4.0.1 or newer
+    - [**gamma_correction**][gamma_correction-url] - Modify gamma of the 2D image.
+      - **Dependencies:**
+        - [OpenCV](https://opencv.org/) version 4.0.1 or newer
 
 ## Instructions
 
@@ -91,3 +94,4 @@ Zivid Samples are distributed under the [BSD license](source/LICENSE).
 [rtde_guide-url]: https://www.universal-robots.com/how-tos-and-faqs/how-to/ur-how-tos/real-time-data-exchange-rtde-guide-22229/
 [downsample-url]: source/applications/advanced/downsample.py
 [create_depth_map-url]: source/applications/advanced/create_depth_map.py
+[gamma_correction-url]: source/applications/advanced/gamma_correction.py
--- a/source/applications/advanced/gamma_correction.py
+++ b/source/applications/advanced/gamma_correction.py
@@ -1,0 +1,92 @@
+"""
+Capture a 2D image and apply a configurable gamma correction.
+"""
+
+import argparse
+import numpy as np
+import cv2
+import zivid
+
+
+def _options():
+    """Configure and take command line arguments from user
+
+    Returns:
+        Arguments from user
+
+    """
+    parser = argparse.ArgumentParser(
+        description=(
+            "Capture 2D image and apply gamma correction\n"
+            "Example:\n\t $ python gamma_correction.py 2"
+        ),
+        formatter_class=argparse.RawTextHelpFormatter,
+    )
+
+    parser.add_argument("gamma", type=float, help="Gamma correction value")
+
+    return parser.parse_args()
+
+
+def adjust_gamma(image, gamma: float):
+    """Adjust gamma on image.
+
+    Copied from : https://www.pyimagesearch.com/2015/10/05/opencv-gamma-correction/
+
+    Args:
+        image: OpenCV image to be corrected
+        gamma: Gamma setting to be applied
+
+    Returns:
+        Gamma adjusted (OpenCV) image
+
+    """
+    # build a lookup table mapping the pixel values [0, 255] to
+    # their adjusted gamma values
+    inv_gamma = 1.0 / gamma
+    table = np.array(
+        [((i / 255.0) ** inv_gamma) * 255 for i in np.arange(0, 256)]
+    ).astype("uint8")
+    # apply gamma correction using the lookup table
+    return cv2.LUT(image, table)
+
+
+def _capture_2d_image(camera):
+    """Capture and extract 2D image, then convert from RGBA and return BGR.
+
+    Args:
+        camera: Zivid Camera handle
+
+    Returns:
+        OpenCV BGR image
+
+    """
+    settings_2d = zivid.Settings2D()
+    with camera.capture_2d(settings_2d) as frame_2d:
+        image = frame_2d.image()
+        image_array = image.to_array()
+        return np.dstack([image_array["b"], image_array["g"], image_array["r"]])
+
+
+def _main():
+    app = zivid.Application()
+
+    user_options = _options()
+
+    camera = app.connect_camera()
+
+    bgr_original = _capture_2d_image(camera)
+    cv2.imwrite("original.jpg", bgr_original)
+
+    print(f"Applying gamma correction: {user_options.gamma}")
+    bgr_adjusted = adjust_gamma(bgr_original, user_options.gamma)
+    cv2.imwrite("adjusted.jpg", bgr_adjusted)
+
+    width = (int)(bgr_original.shape[1] / 2)
+    combined_image = np.hstack([bgr_original[:, :width], bgr_adjusted[:, -width:]])
+    cv2.imshow("Original on left, adjusted on right", combined_image)
+    cv2.waitKey(0)
+
+
+if __name__ == "__main__":
+    _main()


source/applications/advanced/gamma_correction.py
```diff
@@ -0,0 +1,42 @@
...
+    invGamma = 1.0 / gamma
+    table = np.array(
+        [((i / 255.0) ** invGamma) * 255 for i in np.arange(0, 256)]
+    ).astype("uint8")
+    # apply gamma correction using the lookup table
+    return cv2.LUT(image, table)
+
+
+def _load_zivid_image():
+    app = zivid.Application()
```
torbsorb: We usually initialize zivid application in main(). Consider moving this line to beginning of main()

torbsorb: Ok, I will move it!

README.md
```diff
@@ -45,6 +45,9 @@ There are two main categories of samples: **camera** and **applications**. The s
     - [**create_depth_map**][create_depth_map-url] - Import a ZDF point cloud and convert it to OpenCV format, then extract and visualize depth map.
       - **Dependencies:**
         - [OpenCV](https://opencv.org/) version 4.0.1 or newer
+    - [**gamma_correction**][gamma_correction-url] - Show how to modify gamma of the 2D image.
```
SatjaSivcev, line 48: Keep just 'Modify gamma of the 2D image'? (every sample is 'show how to')

source/applications/advanced/gamma_correction.py
```diff
@@ -0,0 +1,100 @@
...
+    for gamma in range(2, 30, 5):
+        print(f"Applying gamma correction: {gamma / 10}")
+        adjusted = adjust_gamma(original, gamma=gamma / 10)
+        cv2.imwrite("adjusted.jpg", adjusted)
+        cv2.imshow("Images", np.hstack([adjusted]))
+        cv2.waitKey(500)
+
+
+if __name__ == "__main__":
+    # If running the script from Spyder IDE, first run '%gui qt'
```
SatjaSivcev, line 99: Is this a copy-paste result or intended?

source/applications/advanced/gamma_correction.py
```diff
@@ -0,0 +1,100 @@
...
+    with camera.capture_2d(settings_2d) as frame_2d:
+        image = frame_2d.image()
+        image.save("original.png")
+        return image.to_array()
+
+
+def _main():
+    app = zivid.Application()
+
+    user_options = _options()
```
SatjaSivcev, line 73: It is unclear what this does. The function description also doesn't help much.

source/applications/advanced/gamma_correction.py
```diff
@@ -0,0 +1,100 @@
...
+def _options():
+    """Configure argparse.
+
+    Returns:
+        argparser arguments
+
+    """
+    parser = argparse.ArgumentParser(
+        description="Capture 2D image and apply gamma correction\
+            \nExample:\n\t $ python gamma_correction.py 2",
```
trym-b, line 21: If you want to do string concatenation you can do it like this (Note: no comma):
```
(
    "Capture 2D image and apply gamma correction\n"
    "Example:\n\t $ python gamma_correction.py 2"
)
```

source/applications/advanced/gamma_correction.py
```diff
@@ -0,0 +1,100 @@
...
+        formatter_class=argparse.RawTextHelpFormatter,
+    )
+
+    parser.add_argument(
+        "gamma", nargs="?", type=float, help="Gamma correction value", default="2"
```
trym-b, line 26: Why have a default value? Is this the value most users would normally use? Would it make sense to make the user always input what correction value they want?

source/applications/advanced/gamma_correction.py
```diff
@@ -0,0 +1,100 @@
...
+    )
+
+    return parser.parse_args()
+
+
+def adjust_gamma(image, gamma=1.0):
```
trym-b, line 32: Does it makes sense to have a default value here? Wouldn't it be better to always require two arguments?

source/applications/advanced/gamma_correction.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def _main():
+    app = zivid.Application()
+
+    user_options = _options()
+
+    if not app.cameras():
+        print(f"No camera connected, will use file camera instead")
+        camera = app.create_file_camera(Path(__file__).parents[2] / "MiscObjects.zdf")
```
trym-b, line 77: I think this should fail unless the user adds an argument `--file-camera` or something like that

source/applications/advanced/gamma_correction.py
```diff
@@ -0,0 +1,100 @@
...
+    else:
+        camera = app.connect_camera()
```
SatjaSivcev, line 79: This is a nice feature for someone with no access to a camera to test a code sample.

Is it really needed though? Is it likely that a potential customer who doesn't have a camera gets this deep into samples, or an existing customer that is doing some testing without a camera? Should we then have this for all samples that can be used with a file camera? If not all, how to decide which ones and based on what criteria (is this one a good candidate for it?). It is a bit of distraction when reading the code, especially with the options part.

I am not saying to remove it but I want to see what we are achieving with it before we decide to have it.

source/applications/advanced/gamma_correction.py
```diff
@@ -0,0 +1,100 @@
...
+    inv_gamma = 1.0 / gamma
+    table = np.array(
+        [((i / 255.0) ** inv_gamma) * 255 for i in np.arange(0, 256)]
+    ).astype("uint8")
+    # apply gamma correction using the lookup table
+    return cv2.LUT(image, table)
+
+
+def _capture_2d_image(camera):
+    """Capture from file, and extract 2D image.
```
SatjaSivcev, line 54: from file? It can also be a regular camera right?

source/applications/advanced/gamma_correction.py
```diff
@@ -0,0 +1,100 @@
...
+    user_options = _options()
+
+    if not app.cameras():
+        print(f"No camera connected, will use file camera instead")
+        camera = app.create_file_camera(Path(__file__).parents[2] / "MiscObjects.zdf")
+    else:
+        camera = app.connect_camera()
+
+    image = _capture_2d_image(camera)
+    original = np.dstack([image["b"], image["g"], image["r"]])
```
SatjaSivcev, line 82: I would argue that this should be inside the _capture_2d_image.
In other samples, we call this kind of image rgb. Here, I think we should call it bgr_original.

source/applications/advanced/gamma_correction.py
trym-b, line 26: Would it make sense to only accept one argument here instead of using `nargs="?"` ?

source/applications/advanced/gamma_correction.py
```diff
@@ -0,0 +1,100 @@
...
+
+    print(f"Applying gamma correction: {user_options.gamma}")
+    adjusted = adjust_gamma(original, gamma=user_options.gamma)
+    cv2.imwrite("adjusted.jpg", adjusted)
+    cv2.imshow("Images", np.hstack([adjusted]))
+    cv2.waitKey(0)
+
+    for gamma in range(2, 30, 5):
+        print(f"Applying gamma correction: {gamma / 10}")
+        adjusted = adjust_gamma(original, gamma=gamma / 10)
```
trym-b, line 92: Why do you divide by 10?

source/applications/advanced/gamma_correction.py
```diff
@@ -0,0 +1,100 @@
...
+        print(f"No camera connected, will use file camera instead")
+        camera = app.create_file_camera(Path(__file__).parents[2] / "MiscObjects.zdf")
+    else:
+        camera = app.connect_camera()
+
+    image = _capture_2d_image(camera)
+    original = np.dstack([image["b"], image["g"], image["r"]])
+
+    print(f"Applying gamma correction: {user_options.gamma}")
+    adjusted = adjust_gamma(original, gamma=user_options.gamma)
```
SatjaSivcev, line 85: Accordingly to the previous comment, this would be bgr_adjusted. It is good to emphesize the weird openCV order b-g-r.

source/applications/advanced/gamma_correction.py
```diff
@@ -0,0 +1,100 @@
...
+    cv2.imwrite("adjusted.jpg", adjusted)
+    cv2.imshow("Images", np.hstack([adjusted]))
+    cv2.waitKey(0)
+
+    for gamma in range(2, 30, 5):
```
trym-b, line 90: I am not sure why you loop over this many values, shouldn't the values be specified by the user?

source/applications/advanced/gamma_correction.py
```diff
@@ -0,0 +1,100 @@
...
+        camera: Zivid Camera handle
+
+    Returns:
+        RGBA image as numpy array
+
+    """
+    settings_2d = zivid.Settings2D()
+    with camera.capture_2d(settings_2d) as frame_2d:
+        image = frame_2d.image()
+        image.save("original.png")
```
SatjaSivcev, line 66: This should probably be out of this function?

SatjaSivcev: And then use imwrite as you do for the other image.

source/applications/advanced/gamma_correction.py
```diff
@@ -0,0 +1,100 @@
...
+        camera = app.create_file_camera(Path(__file__).parents[2] / "MiscObjects.zdf")
+    else:
+        camera = app.connect_camera()
+
+    image = _capture_2d_image(camera)
+    original = np.dstack([image["b"], image["g"], image["r"]])
+
+    print(f"Applying gamma correction: {user_options.gamma}")
+    adjusted = adjust_gamma(original, gamma=user_options.gamma)
+    cv2.imwrite("adjusted.jpg", adjusted)
```
SatjaSivcev, line 86: The other image is png. Make them consistent?

source/applications/advanced/gamma_correction.py
torbsorb: Yes, copy-error

source/applications/advanced/gamma_correction.py
torbsorb: Same function name as we use in convert_zdf.py and universal_robots_perform_hand_eye_calibration.py. I can update the docstring at least

source/applications/advanced/gamma_correction.py
torbsorb: I agree with your reasoning. I didn't want to change too much from the original proposed version... However, if we agree only to use an actual camera I'm totally fine with that.

source/applications/advanced/gamma_correction.py
torbsorb: Yeah, I debated that with myself. Was thinking the function should be "Zivid only", and thus conversion to OpenCV happen outside of the function. If that's not important then we can definitively move it inside.

source/applications/advanced/gamma_correction.py
torbsorb: I'm modifying according to your comment. However, why was it wrong it first place? None of our linters caught it. Is it because you don't like to place `\n` before a line?

source/applications/advanced/gamma_correction.py
torbsorb: @SatjaSivcev 
Again, I didn't want to modify the nature of the original example. I actually agree with you Trym, and I would maybe propose that we comment on the expected values instead of a default.

```python
parser.add_argument(
    "gamma", type=float, help="G < 1 makes image darker, while G > 1 makes image lighter"
)
```

source/applications/advanced/gamma_correction.py
torbsorb: removing file camera from sample

source/applications/advanced/gamma_correction.py
torbsorb: It was necessary when I wanted to use the default value. When the default value goes away, so does `nargs="?"`

source/applications/advanced/gamma_correction.py
torbsorb: Loop wasn't supposed to be pushed... I only did that for my own understanding (/amusement)

source/applications/advanced/gamma_correction.py
torbsorb: again, didn't want to modify the original proposal too much. Sure, let's change it.

torbsorb:
@trym-b and @SatjaSivcev, I've made updates according to both of yours comments.

I found that `adjust_gamma` is a direct copy from https://www.pyimagesearch.com/2015/10/05/opencv-gamma-correction/... should we obfuscate that, or is it ok that we do a blind copy?

source/applications/advanced/gamma_correction.py
```diff
@@ -0,0 +1,89 @@
...
+
+    return parser.parse_args()
+
+
+def adjust_gamma(image, gamma: float):
+    """Adjust gamma on image.
+
+    Args:
+        image: OpenCV image to be corrected
+        gamma: Gamma setting to apply
```
SatjaSivcev, line 38: Use the same tense - image to correct, settings to apply, or image to be corrected, settings to be applied, I don't know which is correct through.

source/applications/advanced/gamma_correction.py
```diff
@@ -0,0 +1,89 @@
...
+
+    user_options = _options()
+
+    camera = app.connect_camera()
+
+    bgr_original = _capture_2d_image(camera)
+    cv2.imwrite("original.jpg", bgr_original)
+
+    print(f"Applying gamma correction: {user_options.gamma}")
+    bgr_adjusted = adjust_gamma(bgr_original, gamma=user_options.gamma)
```
SatjaSivcev, line 82: Not a comment but a question - what is the benefit of passing an argument as `gamma=user_options.gamma` instead of `user_options.gama`? Does it matter because the function is defined as `def adjust_gamma(image, gamma: float):`? What would be the difference if the function was defined as `def adjust_gamma(image, gamma):`?

source/applications/advanced/gamma_correction.py
```diff
@@ -0,0 +1,89 @@
...
+
+def _main():
+    app = zivid.Application()
+
+    user_options = _options()
+
+    camera = app.connect_camera()
+
+    bgr_original = _capture_2d_image(camera)
+    cv2.imwrite("original.jpg", bgr_original)
```
SatjaSivcev, line 79: I would display both images, before and after:
`    cv2.imshow("Images", np.hstack([bgr_original]))
    cv2.waitKey(0)`

SatjaSivcev: And add a title to it, or print out displaying this image.

SatjaSivcev: Same for the one afte gamma.

torbsorb: yeah, I agree. I even thought about splitting the image half way, and show half the image original and half corrected. Won't do it if the code is too complicated

source/applications/advanced/gamma_correction.py
torbsorb: `gamma=` is required if there are default arguments in a function call, and you're using some defaults, but not all. I guess this is Python's overloading

torbsorb: since we removed the default value we should probably just call it with positional arguments.

source/applications/advanced/gamma_correction.py
torbsorb: done, kept 'to be'

source/applications/advanced/gamma_correction.py
trym-b: I reacted to using a `\` to extend a line in python. I think the code is easier to read if we split the string, but not the logic.
I feel `\` is more of a legacy command, like you would using when creating `bash` scripts
The `\n` are fine, it is just the `\` that I have a problem with

PR #61, opened 2020-04-03
Add color balance sample

SatjaSivcev:


martvald:
I feel like 250 lines to do color calibration is just confusing the constumer?

SatjaSivcev:
> 
> 
> I feel like 250 lines to do color calibration is just confusing the costumer?

You are right that the sample is long. I wanted to make the 'main' easy to understand and high level. With a lot of functions I believe the customer doesn't have to read every single line of code, but just the core, which is in the calibration function. All other functions in the main should be clear from their name and are not important in this sample (which is why they are hidden behind a function).

source/applications/advanced/color_balance.py
```diff
@@ -0,0 +1,257 @@
...
+    ]
+
+    mean_red = np.mean(np.reshape(red, -1, []))
+    mean_green = np.mean(np.reshape(green, -1, []))
+    mean_blue = np.mean(np.reshape(blue, -1, []))
+
+    return mean_red, mean_green, mean_blue
+
+
+def _rgb_from_3d_capture(camera):
```
martvald, line 66: The input of this function should be a frame, not camera. It's confusing that you call this function and it will take the image for you.

source/applications/advanced/color_balance.py
```diff
@@ -0,0 +1,257 @@
...
+
+    """
+    print(f"Starting color balance calibration")
+    default_red_balance = camera.settings.red_balance
+    default_blue_balance = camera.settings.blue_balance
+    with camera.update_settings() as updater:
+        updater.settings.blue_balance = 1.0
+        updater.settings.red_balance = 1.0
+
+    rgb = _rgb_from_3d_capture(camera)
```
martvald, line 101: I guess it should be possible to do color calibration for 2D as well? I  understand that this works well due to being able to set rgain and bgain in 3D. But wasn't the main problem 2D?

source/applications/advanced/color_balance.py
```diff
@@ -0,0 +1,257 @@
...
+        f"Mean color values: R =",
+        int(mean_red),
+        f"G =",
+        int(mean_green),
+        f"B =",
+        int(mean_blue),
+    )
+    _display_rgb(rgb, "RGB image before color balance (3D)")
+
+    while int(mean_green) != int(mean_red) and int(mean_green) != int(mean_blue):
```
martvald, line 113: Is it neccesary to retake images? I think the function should just have a check that the exposure time is about right. 

You will end up with bad calibration regardless now if the user f.ex has an overexposed image.

source/applications/advanced/color_balance.py
```diff
@@ -0,0 +1,257 @@
...
+        corrected_blue_balance = camera.settings.blue_balance * mean_green / mean_blue
+        with camera.update_settings() as updater:
+            updater.settings.red_balance = corrected_red_balance
+            updater.settings.blue_balance = corrected_blue_balance
+        rgb = _rgb_from_3d_capture(camera)
+        [mean_red, mean_green, mean_blue] = _compute_mean_rgb(rgb, 100)
+        _display_rgb(rgb, "RGB image after color balance (3D)")
+
+    return (
+        corrected_red_balance / default_red_balance,
```
martvald, line 124: This can't be correct? Must be multipy?

source/applications/advanced/color_balance.py
```diff
@@ -0,0 +1,257 @@
...
+    """
+    rgb_zivid = frame_2d.image().to_array()
+    rgb = np.dstack([rgb_zivid["r"], rgb_zivid["g"], rgb_zivid["b"], rgb_zivid["a"]])
+
+    return rgb
+
+
+def _main():
+
+    _display_color_balance_example_results()
```
martvald, line 236: Why is this part of the example?

source/applications/advanced/color_balance.py
```diff
@@ -0,0 +1,257 @@
...
+        red_balance: Red balance
+        blue_balance: Blue balance
+
+    Returns:
+        corrected_image: Image after color balance
+
+    """
+
+    corrected_image = image
+    image[:, :, 0] = corrected_image[:, :, 0] * np.float32(red_balance)
```
martvald, line 212: Be carefult. uint8 * float32 = float32. Make sure this is the behavior you want

source/applications/advanced/color_balance.py
SatjaSivcev: The problem is that a few iterations are necessary (most times 3) to get rgain and bgain that make R G B mean values equal. That is why the calibration is in 3D (to get those rgain and bgain) and correction is in 2D (where you apply rgain and bgain on the image).

source/applications/advanced/color_balance.py
SatjaSivcev: If the image is overexposed that will indeed lead to a bad calibration.

I don't understand the first comment. The reason it is necessary to retake images is that a single iteration is not enough to get rgain and bgain that make R G B channel mean values equal. Or are you suggesting instead to apply gains on the R G B channels, recalculate their mean, compare them, repeat?

Are you suggesting to add a capture loop that finds the right exposure?

source/applications/advanced/color_balance.py
SatjaSivcev: I wanted to demonstrate what the sample does. Perhaps this should be part of the Readme file for this sample.

source/applications/advanced/color_balance.py
SatjaSivcev: I will investigate.

source/applications/advanced/color_balance.py
SatjaSivcev: It is correct in my opinion (and it is proved to work). The point is that this gain is applied to an image here: https://github.com/zivid/python-samples/blob/color-balance/source/applications/advanced/color_balance.py#L212
That image is a result of capture with default_red_balance. That is why you have to divide it with default_red_balance to get rid of it and multiply with the corrected_red_balance because that is what you want.

Maybe it would be more understandable if this function was to return just corrected_red_balance, and then when we apply it we divide by default_red_balance and multiply wiht the corrected_red_balance. However, then we need to save the defaut_red_balance in main and transfer it to _apply_color_balance, or hard code it.

What do you think?

source/applications/advanced/color_balance.py
martvald: But why is one iteration not enought?

source/applications/advanced/color_balance.py
martvald: Ok, so since you can't turn of Rgain and Bgain for 2D capture. You have to multiply image with the corrected settings. While in 3D then new Rgain is Rgain*correctedRgain. Not sure that's the easiest format for it.

source/applications/advanced/color_balance.py
martvald: Yes, I would rather seperate these at least.

source/applications/advanced/color_balance.py
SatjaSivcev: It is nearly enough. the difference in R G B after one iteration is around somewhere between 1 and 3 (0 to 255). I don't understand why this happens though.

source/applications/advanced/color_balance.py
SatjaSivcev: I will remove this function completely.

source/applications/advanced/color_balance.py
SatjaSivcev: I haven't changed this. In average, three iterations are required.

source/applications/advanced/color_balance.py
SatjaSivcev: `corrected_rgb = np.copy(rgb)
    corrected_rgb[:, :, 0] = rgb[:, :, 0] * red_balance / default_red_balance
    corrected_rgb[:, :, 2] = rgb[:, :, 2] * blue_balance / default_blue_balance`

Not changed. Since, corrected_rgb is first created as uint8, the multiplication will return float but will be converted to uint8 as it has to adapt to type of corrected_rgb

source/applications/advanced/color_balance.py
SatjaSivcev: I've removed it from the sample.

SatjaSivcev:
Ready for the next round.

source/applications/advanced/color_balance.py
```diff
@@ -0,0 +1,203 @@
...
+        settings.iris = iris
+        settings.exposure_time = datetime.timedelta(microseconds=exposure_time)
+        settings.brightness = brightness
+        settings.gain = gain
+    if dimension == "2d":
+        settings = zivid.Settings2D()
+        settings.iris = iris
+        settings.exposure_time = datetime.timedelta(microseconds=exposure_time)
+        settings.brightness = brightness
+        settings.gain = gain
```
trym-b, line 52: I would have added an `else: raise ValueError()` here. I would also have used `if-elif-else`

source/applications/advanced/color_balance.py
```diff
@@ -0,0 +1,203 @@
...
+    Returns:
+        mean_red: Red channel mean value
+        mean_green: Green channel mean value
+        mean_blue: Blue channel mean value
+
+    """
+    height = np.shape(rgb)[0]
+    width = np.shape(rgb)[1]
+    red = rgb[
+        int((height - pixels) / 2) : int((height + pixels) / 2),
```
trym-b, line 91: `height - pixels` is calculated 3 times, I would do the computations first, then use the result in order to avoid copy-paste errors

source/applications/advanced/color_balance.py
```diff
@@ -0,0 +1,203 @@
...
+        int((width - pixels) / 2) : int((width + pixels) / 2),
+        1,
+    ]
+    blue = rgb[
+        int((height - pixels) / 2) : int((height + pixels) / 2),
+        int((width - pixels) / 2) : int((width + pixels) / 2),
+        2,
+    ]
+    mean_red = np.mean(np.reshape(red, -1, []), dtype=np.float64)
```
trym-b, line 105: What is the last argument in `np.reshape(red, -1, [])` ?

source/applications/advanced/color_balance.py
```diff
@@ -0,0 +1,203 @@
...
+    mean_green = np.mean(np.reshape(green, -1, []), dtype=np.float64)
+    mean_blue = np.mean(np.reshape(blue, -1, []), dtype=np.float64)
+
+    return mean_red, mean_green, mean_blue
```
trym-b, line 109: What version of python is this sample targeted?

source/applications/advanced/color_balance.py
```diff
@@ -0,0 +1,203 @@
...
+            first_iteration = False
+        [mean_red, mean_green, mean_blue] = _compute_mean_rgb(rgb, 100)
+        print(
+            f"Mean color values: R =",
+            int(mean_red),
+            f"G =",
+            int(mean_green),
+            f"B =",
+            int(mean_blue),
+        )
```
trym-b, line 168: How about something like this:
```
print(
    (
       "Mean color values: R = "
        f"{int(mean_red)}\n"
        "G = "
        f"{int(mean_green)}\n"
        f"B = "
        f"{int(mean_blue)}\n"
    )
)
```
The f-strings in the current code does not do anything, since you are using comma to separate the arguments of the print

source/applications/advanced/color_balance.py
SatjaSivcev: Done

source/applications/advanced/color_balance.py
SatjaSivcev: Done

source/applications/advanced/color_balance.py
SatjaSivcev: It is the order, which doesn't matter.

SatjaSivcev: I can also remove it `np.reshape(red, -1)` it would work the same

trym-b: Is it a required argument? Or could you drop it altogether?

source/applications/advanced/color_balance.py
SatjaSivcev: I guess 3.5 or higher

source/applications/advanced/color_balance.py
SatjaSivcev: Done like this (removed \n and added space so it is in one line):
`        print(
            (
               "Mean color values: R = "
                f"{int(mean_red)} "
                "G = "
                f"{int(mean_green)} "
                f"B = "
                f"{int(mean_blue)} "
            )
        )`

source/applications/advanced/color_balance.py
trym-b: f-strings are python 3.6

SatjaSivcev: All our samples use f-strings. We never specify what version is required in the samples. Out tutorial 'setting up python' actually points at 3.7. Zivid-Python suggests 3.5 though. In any case, we should make an issue for the whole repository and not solve this for this PR.

SatjaSivcev:
Done with your changes @trym-b

source/applications/advanced/color_balance.py
```diff
@@ -0,0 +1,214 @@
...
+        settings.brightness = brightness
+        settings.gain = gain
+    elif dimension == "2d":
+        settings = zivid.Settings2D()
+        settings.iris = iris
+        settings.exposure_time = datetime.timedelta(microseconds=exposure_time)
+        settings.brightness = brightness
+        settings.gain = gain
+    else:
+        raise ValueError("The dimension value should be '3d' or '2d'")
```
trym-b, line 57: You also probably want to write what you got in: `, got: {dimension}`

source/applications/advanced/color_balance.py
trym-b: Sounds good. I think it is a dilemma, at one hand we want to use the latest and greatest from python, to show that we are trying our best to keep up to date and write the best code possible, but at the other hand we would also like samples to work with older python versions.

source/applications/advanced/color_balance.py
```diff
@@ -0,0 +1,214 @@
...
+            _display_rgb(rgb, "RGB image before color balance (3D capture)")
+            first_iteration = False
+        [mean_red, mean_green, mean_blue] = _compute_mean_rgb(rgb, 100)
+        print(
+            (
+                "Mean color values: R = "
+                f"{int(mean_red)} "
+                "G = "
+                f"{int(mean_green)} "
+                f"B = "
```
trym-b, line 176: Minor: This does not need to be an f-string

source/applications/advanced/color_balance.py
trym-b: The reason why I asked this was that we could return all of this information in one block:
```
@dataclass
class MeanColor:
    red: np.Array
    green: np.Array
    blue: np.Array
```
Which you could use like this:
`return MeanColor(red=mean_red, green=mean_green, blue=mean_blue)`

And later use it like this:
`mean_color.red`

source/applications/advanced/color_balance.py
SatjaSivcev: Done.

source/applications/advanced/color_balance.py
SatjaSivcev: Done

source/applications/advanced/color_balance.py
SatjaSivcev: I've implemented this.

SatjaSivcev:
Ready for the next round @trym-b

source/applications/advanced/color_balance.py
```diff
@@ -0,0 +1,239 @@
...
+    Returns:
+        corrected_red_balance: Corrected red balance
+        corrected_blue_balance: Corrected blue balance
+
+    """
+    print(f"Starting color balance calibration")
+    corrected_red_balance = 1.0
+    corrected_blue_balance = 1.0
+    settings_list = []
+    settings_list.append(settings_3d)
```
trym-b, line 177: This can be simplified to: `settings_list = [settings_3d]`

SatjaSivcev: Done

SatjaSivcev:
I rebased, can you approve @trym-b ?

PR #68, opened 2020-05-30
Sample did not show camera information

torbsorb:
Old output:

```shell
Python:       3.7.4
zivid-python: 1.0.0.1.8.1
Zivid SDK:    1.8.1+6967bc1b-1
```

New output:
```shell
Python:       3.7.4
zivid-python: 1.0.0.1.8.1
Zivid SDK:    1.8.1+6967bc1b-1
Camera Info:  { Camera info: { Model Name: Zivid One Plus M, Serial Number: 19200CB2, Hardware Revision: { Major: 3, Minor: 2 }, Firmware Version: 1.6.7, USB Path: na, Internal camera id: 0510737 } }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ There are two main categories of samples: **camera** and **applications**. The s
     - [**capture_hdr**][capture_hdr-url] - This example shows how to acquire HDR images from the Zivid camera.
     - [**capture_hdr_complete_settings**][capture_hdr_complete_settings-url] - This example shows how to acquire HDR images from the Zivid camera and configure all settings.
   - **info_util_other**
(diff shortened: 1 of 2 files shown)


torbsorb:
> 
> 
> LGTM.
> 
> Should we consider upgrading this so that it can be easily imported and used in a other sample scripts?

In what way?

SatjaSivcev:
> 
> 
> > LGTM.
> > Should we consider upgrading this so that it can be easily imported and used in a other sample scripts?
> 
> In what way?

So that this would be possible:

from print_version_info import print_zivid_info
print_zivid_info()

PR #74, opened 2020-07-03
Update color_balance samples

torbsorb:
This PR focus on the changes required due to the color balance updates in SDK 2.0.

### DIFF
--- a/source/applications/advanced/color_balance.py
+++ b/source/applications/advanced/color_balance.py
@@ -4,6 +4,7 @@ Balance color for RGB image.
 
 import datetime
 from dataclasses import dataclass
+import math
 import numpy as np
 import matplotlib.pyplot as plt
 import zivid
@@ -24,6 +25,21 @@ class MeanColor:
     red: np.array
     green: np.array
     blue: np.array
+
+
+def _get_next_aperture_setting(current_aperture: float, stop_increment: float) -> float:
+    """Get next aperture setting.
+
+    Args:
+        current_aperture: Current aperture
+        stop_increment: Exposure stops to move (typically 1/4, 1/2 or 1/1)
+
+    Returns:
+        New aperture
+
+    """
+    # new_aperture = current_aperture * 2 ^ (0.5 * stop_increment)
+    return current_aperture * math.pow(2, 0.5 * stop_increment)
 
 
 def _frame2d_to_rgb(frame: zivid.Frame2D):
@@ -56,15 +72,15 @@ def _display_rgb(rgb, title):
     plt.show(block=False)
 
 
-def _compute_mean_rgb(rgb, pixels):
-    """Compute mean RGB values.
+def _get_center_mask(rgb, pixels):
+    """Get central square RGB values.
 
     Args:
         rgb: RGB image
         pixels: Number of central pixels (^2) for computation
 
     Returns:
-        mean_color: RGB channel mean values
+        masked_rgb: Masked RGB
 
     """
     height = np.shape(rgb)[0]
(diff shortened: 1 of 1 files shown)


source/applications/advanced/color_balance.py
```diff
@@ -64,26 +83,60 @@ def _compute_mean_rgb(rgb, pixels):
...
+    Returns:
+        settings_2d: 2D capture settings without saturation on average
+
+    """
+    settings_2d = zivid.Settings2D(
+        acquisitions=[
+            zivid.Settings2D.Acquisition(
+                aperture=5.66,
+                exposure_time=datetime.timedelta(microseconds=80000),
+                brightness=0.0,
```
SatjaSivcev, line 122: Should we be explicit here on not using the projector?

source/applications/advanced/color_balance.py
```diff
@@ -4,11 +4,25 @@
 
 import datetime
 from dataclasses import dataclass
+import math
 import numpy as np
 import matplotlib.pyplot as plt
 import zivid
 
 
+@dataclass(frozen=True)
```
SatjaSivcev, line 13: This is a nice class but I am not sure it is needed for this sample.

source/applications/advanced/color_balance.py
```diff
@@ -26,6 +40,11 @@ class MeanColor:
     blue: np.array
 
 
+def _get_next_aperture_setting(current_aperture: float, stop: float) -> float:
+    # new_aperture = current_aperture * 2 ^ (0.5 * stop)
```
SatjaSivcev, line 44: Not sure this works. Should be tested.

source/applications/advanced/color_balance.py
```diff
@@ -64,26 +83,60 @@ def _compute_mean_rgb(rgb, pixels):
...
+            settings_2d.acquisitions[0].aperture, StopIncrement.quarter
+        )
+        rgb = _convert_frame2d_to_numpy_float_rgb(camera.capture(settings_2d))
+        rgb_masked = _get_center_mask(rgb, 100)
+        mean_rgb = np.mean(np.mean(rgb_masked, axis=0), axis=0)
+        print(f"RGB mean: {mean_rgb}, Aperture: {settings_2d.acquisitions[0].aperture}")
+    return settings_2d
 
 
 def _color_balance_calibration(camera, settings_2d):
```
SatjaSivcev, line 142: I tried running this sample but it failed - blue out of range

source/applications/advanced/color_balance.py
torbsorb: What do you mean? Add comments?

source/applications/advanced/color_balance.py
torbsorb: I tested it, pretty sure it works.

source/applications/advanced/color_balance.py
SatjaSivcev: I like the fixup.

source/applications/advanced/color_balance.py
SatjaSivcev: No, I meant that we maybe should not hard code it to 0 but something else. It could be that someone is using the camera in e.g. strong yellow light with the projector. Cases like that may be more common than those where the projector is off.

torbsorb: Maybe highlight these settings in main(), making a point that these should be set manually? I also thought about whether to update the automatic setting function to limit the Aperture and use Exposure Time instead and maybe even Gain.

PR #76, opened 2020-07-03
Support sdk2 update hand eye

torbsorb:
This PR is the first in the sequence where we expect CI to be green. Update to SDK 2.0.0 should now be completed

### DIFF
--- a/source/applications/advanced/hand_eye_calibration/calibrate_eye_to_hand.py
+++ b/source/applications/advanced/hand_eye_calibration/calibrate_eye_to_hand.py
@@ -3,7 +3,7 @@ import datetime
 import code
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


source/applications/advanced/color_balance.py
```diff
@@ -154,7 +154,6 @@ def _color_balance_calibration(camera, settings_2d):
     print("Starting color balance calibration")
     corrected_red_balance = 1.0
     corrected_blue_balance = 1.0
-    # TODO: Update algorithm to account for green as well.
```
SatjaSivcev, line 157: this should be in the other PR

torbsorb: it is, note that all these PRs are following each other

PR #77, opened 2020-07-03
Updated path to point cloud example in readme

ptrks:
URL mapping didn't exist for the point cloud example.


SatjaSivcev:
This will be soon fixed in the next PR. Thanks for your input!

PR #79, opened 2020-07-08
Change 'contrast' to 'snr'

SatjaSivcev:


SatjaSivcev:
This doesn't work yet as 'snr' is not yet implemented in zivid-python

PR #80, opened 2020-08-10
Update values for acquisitions

ajlandau:
Update of settings variables for acquisitions to match other samples' settings.

### DIFF
--- a/source/camera/basic/capture.py
+++ b/source/camera/basic/capture.py
@@ -8,9 +8,10 @@ def _main():
     camera = app.connect_camera()
 
     with camera.update_settings() as updater:
-        updater.settings.iris = 40
-        updater.settings.exposure_time = datetime.timedelta(microseconds=40000)
-        updater.settings.filters.reflection.enabled = True
+        updater.settings.iris = 20
+        updater.settings.exposure_time = datetime.timedelta(microseconds=6500)
+        updater.settings.filters.outlier.enabled(True)
+        updater.settings.filters.outlier.threshold(5)
 
     with camera.capture() as frame:
         frame.save("Result.zdf")
--- a/source/camera/basic/capture_2d.py
+++ b/source/camera/basic/capture_2d.py
@@ -8,8 +8,10 @@ def _main():
     camera = app.connect_camera()
 
     settings_2d = zivid.Settings2D()
-    settings_2d.iris = 50
-    settings_2d.exposure_time = datetime.timedelta(microseconds=50000)
+    settings_2d.iris = 20
+    settings_2d.exposure_time = datetime.timedelta(microseconds=30000)
+    settings_2d.brightness(1.8)
+    settings_2d.gain(2.0)
 
     with camera.capture_2d(settings_2d) as frame_2d:
         image = frame_2d.image()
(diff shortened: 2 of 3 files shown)


continuous-integration/requirements.txt
```diff
@@ -1,4 +1,5 @@
 black
 flake8
 pylint
-darglint
\ No newline at end of file
+darglint
+scikit-build
```
SatjaSivcev, line 5: What do we need this one for?

requirements.txt
```diff
@@ -3,6 +3,6 @@ netCDF4
 matplotlib
 vtk
 vtk_visualizer
-opencv-python
+opencv-python==4.2.0.34
```
SatjaSivcev, line 6: This should be a separate commit message with good explanation.

source/camera/basic/capture.py
```diff
@@ -8,9 +8,10 @@ def _main():
     camera = app.connect_camera()
 
     with camera.update_settings() as updater:
-        updater.settings.iris = 40
-        updater.settings.exposure_time = datetime.timedelta(microseconds=40000)
-        updater.settings.filters.reflection.enabled = True
+        updater.settings.iris = 11.31
```
SatjaSivcev, line 11: The iris is different here compared to the cpp sample. Typo?

https://github.com/zivid/zivid-cpp-samples/blob/master/source/Camera/Basic/Capture/Capture.cpp

SatjaSivcev: The iris values are available in the porting guide on the KB.

source/camera/basic/capture_hdr_complete_settings.py
```diff
@@ -17,55 +17,28 @@ def _main():
...
+    for i in range(len(settings_collection)):
+        settings_collection[i].iris = iris_setting[i]
+        settings_collection[i].exposure_time = datetime.timedelta(
+            microseconds=exposure_setting[i]
+        )
+        settings_collection[i].brightness = 1
+        settings_collection[i].gain = gain_setting[i]
+        settings_collection[i].bidirectional = 0
+        settings_collection[i].filters.contrast.enabled = True
+        settings_collection[i].filters.Contrast.threshold = 0.5
```
SatjaSivcev, line 33: this should be 5?

source/camera/basic/capture_2d.py
```diff
@@ -8,8 +8,10 @@ def _main():
     camera = app.connect_camera()
 
     settings_2d = zivid.Settings2D()
-    settings_2d.iris = 50
-    settings_2d.exposure_time = datetime.timedelta(microseconds=50000)
+    settings_2d.iris = 11.31
```
SatjaSivcev, line 11: Zivid-python is based on 1.8 SDK. Therefore we can't use f-numbers but appropriate iris values. This is true for all samples.

continuous-integration/requirements.txt
ajlandau: Removed

requirements.txt
ajlandau: This issue was found and fixed in a new PR, approved by Trym.

source/camera/basic/capture_2d.py
ajlandau: Fixed

ajlandau: Fixed

ajlandau:
All comments addressed, rebased and updated the commit message. Let me know if any other changes are needed.

PR #90, opened 2020-09-27
Update Hand-Eye samples for new API

torbsorb:


eskaur:
Seems fine to me. Have you tried running it?

I have some comments on how you set Settings, but it seems like that belongs in another PR.

eskaur:
Did you try running it? @SatjaSivcev

PR #99, opened 2021-01-03
Port to Github Actions

eskaur:
This commit changes the CI from using Azure Pipelines to using Github
Actions. What happens inside the Docker testing-containers is completely
unchanged.


.github/workflows/main.yml
```diff
@@ -0,0 +1,27 @@
+name: Main CI workflow
+# Note: If the workflow name is changed, the CI badge URL in the README must also be updated
+
+on:
+  push:       # Push trigger runs on any pushed branch.
+  schedule:   # Scheduled trigger runs on the latest commit on the default branch.
+    - cron:  '0 22 * * *'
```
knatten, line 7: The trigger time has changed, but I assume that doesn't matter?

.github/workflows/main.yml
```diff
@@ -0,0 +1,27 @@
...
+  Linux:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: ["ubuntu:18.04"]
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+        with:
+          lfs: true
```
knatten, line 19: Do we use lfs in this repo? Or is it just to be sure in case we start using it later?

SatjaSivcev: We don't use it (assuming lfs here stands for https://git-lfs.github.com/) and we don't plan to, to the best of my knowledge, so I don't see the reason to have it, but it doesn't cost us anything to keep it.
If we were to use lfs, users would have to set it up on their PC to be able to get large files. We thought it was simpler that they just downloads SampleData as a zip file instead (instructions are in README).

.github/workflows/main.yml
eskaur: Yeah it used to be "5 minutes past the hour", but it seemed pointless so I just simplified it to "on the hour".

.github/workflows/main.yml
eskaur: Good catch, this was a mistake. It's a leftover from me using the zivid-python repo as inspiration (which does use LFS). 
I'll remove it.

knatten:
Review done

.github/workflows/main.yml
eskaur: Removed.

eskaur:
Ready for next round.

SatjaSivcev (dismissed):
LGTM. I suggested that we should add 16.04, 20.04, and Windows since we want to check installing requirements (python packages), but we can do that in a separate PR.

SatjaSivcev (approved):
LGTM. I suggested that we should add 16.04, 20.04, and Windows since we want to check installing requirements (python packages), but we can do that in a separate PR.

PR #110, opened 2021-02-09
Replace pptk visualizer with open3D

chrisasc:
Replacing PPTK with Open3D visualizer, since PPTK does not support Python 3.8. 
Open3D supports all python versions from 3.5 and is updated on a regular basis. 

I'm planning on updating all our visualization samples with the same function. but before I do that I want to make sure that the _display_pointcloud() function is good enough.

### DIFF
--- a/source/applications/basic/visualization/read_zdf_vis_3d.py
+++ b/source/applications/basic/visualization/read_zdf_vis_3d.py
@@ -4,12 +4,11 @@ This example shows how to read point cloud data from a ZDF file and visualize it
 The ZDF file for this sample can be found under the main instructions for Zivid samples.
 """
 
-import math
 from pathlib import Path
 import numpy as np
 import matplotlib.pyplot as plt
-import pptk
 import zivid
+import open3d as o3d
 
 from sample_utils.paths import get_sample_data_path
 
@@ -50,36 +49,8 @@ def _display_depthmap(xyz):
     plt.show(block=False)
 
 
-def _get_mid_point(xyz):
-    """Calculate mid point from average of the 100 centermost points.
-
-    Args:
-        xyz: X, Y and Z images (point cloud co-ordinates)
-
-    Returns:
-        mid_point: Calculated mid point
-
-    """
-    xyz_center_cube = xyz[
-        int(xyz.shape[0] / 2 - 5) : int(xyz.shape[0] / 2 + 5),
-        int(xyz.shape[1] / 2 - 5) : int(xyz.shape[1] / 2 + 5),
-        :,
-    ]
-    return (
-        np.nanmedian(xyz_center_cube[:, :, 0]),
-        np.nanmedian(xyz_center_cube[:, :, 1]),
-        np.nanmedian(xyz_center_cube[:, :, 2]),
-    )
-
-
-def _display_pointcloud(rgb, xyz):
-    """Display point cloud.
-
-    Display the provided point cloud `xyz`, and color it with `rgb`.
-
-    We take the centermost co-ordinate as 'lookat' point. We assume that
-    camera location is at azimuth -pi/2 and elevation -pi/2 relative to
-    the 'lookat' point.
+def _display_pointcloud(xyz, rgb):
+    """Display point cloud provided from 'xyz' with colors from 'rgb'.
 
     Args:
         rgb: RGB image
@@ -88,13 +59,24 @@ def _display_pointcloud(rgb, xyz):
     Returns None
 
     """
-    mid_point = _get_mid_point(xyz)
-    point_cloud_to_view = xyz
-    point_cloud_to_view[np.isnan(xyz[:, :, 2])] = 0
-    viewer = pptk.viewer(point_cloud_to_view)
-    viewer.attributes(rgb.reshape(-1, 3) / 255.0)
-    viewer.set(lookat=mid_point)
-    viewer.set(phi=-math.pi / 2, theta=-math.pi / 2, r=mid_point[2])
+    xyz = np.nan_to_num(xyz).reshape(-1, 3)
+    rgb = rgb.reshape(-1, 3)
+
+    point_cloud_open3d = o3d.geometry.PointCloud(o3d.utility.Vector3dVector(xyz))
+    point_cloud_open3d.colors = o3d.utility.Vector3dVector(rgb / 255)
+
+    visualizer = o3d.visualization.Visualizer()  # pylint: disable=no-member
+    visualizer.create_window()
+    visualizer.add_geometry(point_cloud_open3d)
+
+    visualizer.get_render_option().background_color = (0, 0, 0)
+    visualizer.get_render_option().point_size = 1
+    visualizer.get_render_option().show_coordinate_frame = True
+    visualizer.get_view_control().set_front([0, 0, -1])
+    visualizer.get_view_control().set_up([0, -1, 0])
+
+    visualizer.run()
+    visualizer.destroy_window()
 
 
 def _main():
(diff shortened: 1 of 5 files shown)


source/applications/basic/visualization/read_zdf_vis_3d.py
```diff
@@ -50,51 +49,37 @@ def _display_depthmap(xyz):
...
-    viewer.set(phi=-math.pi / 2, theta=-math.pi / 2, r=mid_point[2])
+    point_cloud_to_view = np.nan_to_num(xyz)
+    point_cloud_to_view = point_cloud_to_view.reshape(-1, 3)
+    rgb = rgb.reshape(-1, 3)
+
+    pcd = o3d.geometry.PointCloud(o3d.utility.Vector3dVector(xyz))
+    pcd.colors = o3d.utility.Vector3dVector(rgb / 255)
+
+    visualizer = o3d.visualization.Visualizer()
+    visualizer.create_window(window_name="Open3D Zivid visualizer")
```
SatjaSivcev, line 72: Do we need 'Zivid' here?

source/applications/basic/visualization/read_zdf_vis_3d.py
```diff
@@ -50,51 +49,37 @@ def _display_depthmap(xyz):
...
-    point_cloud_to_view[np.isnan(xyz[:, :, 2])] = 0
-    viewer = pptk.viewer(point_cloud_to_view)
-    viewer.attributes(rgb.reshape(-1, 3) / 255.0)
-    viewer.set(lookat=mid_point)
-    viewer.set(phi=-math.pi / 2, theta=-math.pi / 2, r=mid_point[2])
+    point_cloud_to_view = np.nan_to_num(xyz)
+    point_cloud_to_view = point_cloud_to_view.reshape(-1, 3)
+    rgb = rgb.reshape(-1, 3)
+
+    pcd = o3d.geometry.PointCloud(o3d.utility.Vector3dVector(xyz))
```
SatjaSivcev, line 68: What do you do here? Why 'pcd'?

source/applications/basic/visualization/read_zdf_vis_3d.py
```diff
@@ -50,51 +49,37 @@ def _display_depthmap(xyz):
...
 
     """
-    mid_point = _get_mid_point(xyz)
-    point_cloud_to_view = xyz
-    point_cloud_to_view[np.isnan(xyz[:, :, 2])] = 0
-    viewer = pptk.viewer(point_cloud_to_view)
-    viewer.attributes(rgb.reshape(-1, 3) / 255.0)
-    viewer.set(lookat=mid_point)
-    viewer.set(phi=-math.pi / 2, theta=-math.pi / 2, r=mid_point[2])
+    point_cloud_to_view = np.nan_to_num(xyz)
```
SatjaSivcev, line 64: Does this create fake points in the point cloud?

SatjaSivcev: Why 'point_cloud_to_view'?

robertkt1: Do we need to remove the nan points? or can open3d handle the nans?

source/applications/basic/visualization/read_zdf_vis_3d.py
chrisasc: Nope, i can remove the naming here. This is maybe better so we don't confuse customers to believe we have our own viewer.

source/applications/basic/visualization/read_zdf_vis_3d.py
chrisasc: I forgot to change that to a proper name, my bad! This is now point_cloud_to_view

source/applications/basic/visualization/read_zdf_vis_3d.py
chrisasc: We need to remove them otherwise we can't see a point cloud

source/applications/basic/visualization/read_zdf_vis_3d.py
chrisasc: Changed it to point_cloud_open3d as that is what we do in C++, e.g. pointCloudPCL.

source/applications/basic/visualization/read_zdf_vis_3d.py
SatjaSivcev: Exactly. I agree that it is better to remove it.

source/applications/basic/visualization/read_zdf_vis_3d.py
SatjaSivcev: Then we need to align on what we want to do. By default it will replace nans to zeros. the downside is that if you zoom to (0,0,0) you will probably see a point (many points stacked one on top of each other). Alternative is to set nans to a value we like (it could be a very large positive or negative number). I am not saying this is better, since it has the same downside (if you zoom in there you will see the point(s)). What is your opinion on this @chrisasc @robertkt1? Anything I missed? I guess nan to zero is common and ok to do?

Reference: https://numpy.org/doc/stable/reference/generated/numpy.nan_to_num.html

SatjaSivcev (dismissed):
I am happy with the code now. We just have to agree on what we do with 'np.nan_to_num(xyz)' - after that you can merge and go ahead modifying rest of visualizations.

chrisasc:
> I am happy with the code now. We just have to agree on what we do with 'np.nan_to_num(xyz)' - after that you can merge and go ahead modifying rest of visualizations.

Perfect. What do you mean with "agree on what we do with 'np.nan_to_num(xyz)'"? We need to convert from NaN to 0. Previously we did this: 
`point_cloud_to_view[np.isnan(xyz[:, :, 2])] = 0`
But I think that the np.nan_to_numb() is a nicer way of doing this.

SatjaSivcev:
> > I am happy with the code now. We just have to agree on what we do with 'np.nan_to_num(xyz)' - after that you can merge and go ahead modifying rest of visualizations.
> 
> Perfect. What do you mean with "agree on what we do with 'np.nan_to_num(xyz)'"? We need to convert from NaN to 0. Previously we did this:
> `point_cloud_to_view[np.isnan(xyz[:, :, 2])] = 0`
> But I think that the np.nan_to_numb() is a nicer way of doing this.

I meant this: https://github.com/zivid/zivid-python-samples/pull/110#discussion_r574364214
But if this is what we have already been doing, then I am ok, you can merge.

source/applications/basic/visualization/read_zdf_vis_3d.py
chrisasc: One possibility is to remove all NaN points, but this would make the point cloud unorganized. I think that converting them to (0,0,0) is the best option as this is what's common to do (it happens in Halcon, we do it in C++, and I believe that's what most open-source libraries do). 

Unless @robertkt1 has a good argument for not converting to (0,0,0), I will keep doing this.

chrisasc:
Updated README and dependencies.txt too

source/applications/basic/visualization/read_zdf_vis_3d.py
```diff
@@ -114,9 +96,7 @@ def _main():
 
     _display_depthmap(xyz)
 
-    _display_pointcloud(rgba[:, :, 0:3], xyz)
-
-    input("Press Enter to close...")
```
robertkt1, line 119: Why do we remove this line in this sample but not other samples?

robertkt1: Line 119

robertkt1:
Is there a reason we chose to copy xyz and rgb outside of _display_pointcloud?

source/applications/basic/visualization/read_zdf_vis_3d.py
chrisasc: Thought it was not necessary, but I understand why we have it there now! I put it back in

robertkt1 (approved):
LGTM

PR #116, opened 2021-07-28
Add warm up sample

ajlandau:


source/camera/info_util_other/warm-up-sample.py
```diff
@@ -0,0 +1,54 @@
...
+    capture_cycle = timedelta(seconds=5)
+    max_capture_time = timedelta(milliseconds=1000)
+
+    print("Getting camera settings")
+    suggest_settings_parameters = zivid.capture_assistant.SuggestSettingsParameters(
+        max_capture_time=max_capture_time,
+        ambient_light_frequency=zivid.capture_assistant.SuggestSettingsParameters.AmbientLightFrequency.none,
+    )
+    settings = zivid.capture_assistant.suggest_settings(camera, suggest_settings_parameters)
+
```
robertkt1, line 26: In C++ we are printing How the warm up time. Should be consistent

source/camera/info_util_other/warm-up-sample.py
```diff
@@ -0,0 +1,54 @@
...
+    before_warmup = datetime.now()
+
+    print("Starting warm up")
```
robertkt1, line 29: An we are not printing "Starting warm up" in c++. Should be consistent. 
Maybe this and the print above can be merged if we want to print both. 
Starting warm up, total x minutes?
If we only want to keep one I think "starting" is nice than the 10 minutes. But no strong opinions here

source/camera/info_util_other/warm-up-sample.py
```diff
@@ -0,0 +1,54 @@
...
+            sleep(capture_cycle.seconds - duration.seconds)
+        else:
+            print(
+                "Your capture time is longer than your desired capture cycle. Please increase the desired capture cycle."
+            )
+
+        remaining_time = warmup_time - (datetime.now() - before_warmup)
+        remaining_time_minutes = (remaining_time.seconds % 3600) // 60
+        remaining_time_seconds = remaining_time.seconds % 60
+        print(f"Remaining time is {remaining_time_minutes} minutes and {remaining_time_seconds} seconds.")
```
robertkt1, line 48: Since we are not printing number of hours, the remaining minutes will be wrong if the time is above 60 minutes?

robertkt1 (dismissed):
Not sure about the CI error. But the sample LGTM

robertkt1:
I take back the approval: see error in CI
image
snake_case module name, and shorten the printout line.

robertkt1 (dismissed):
LGTM

robertkt1 (approved):

PR #124, opened 2021-11-11
Update capture_hdr_complete_settings.py

SatjaSivcev:
This commit modifies the code sample to make it more similar to its
C++ and C# counterparts:

https://github.com/zivid/zivid-cpp-samples/blob/master/source/Camera/Basic/CaptureHDRCompleteSettings/CaptureHDRCompleteSettings.cpp

https://github.com/zivid/zivid-csharp-samples/blob/master/source/Camera/Basic/CaptureHDRCompleteSettings/CaptureHDRCompleteSettings.cs

### DIFF
--- a/source/camera/basic/capture_hdr_complete_settings.py
+++ b/source/camera/basic/capture_hdr_complete_settings.py
@@ -10,6 +10,25 @@ import datetime
 import zivid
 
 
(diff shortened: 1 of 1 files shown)


torbsorb (approved):
very good!

PR #165, opened 2023-08-24
change right indent

honeybadger78:
There was a python indent issue.


andersfagerli:
Hi!

Thanks for taking the time for making the PR. Our process of releasing these code samples is to make them in an internal repository (not public), and then a PR to these public repositories automatically gets created where we finally merge them. We're currently adding this change to the Python sample, but it's taking a bit more time than expected.

This should hopefully be fixed within the week, but it will happen in another PR.

PR #170, opened 2023-12-13
Link to zivid-python in README.md for supported Python versions

johningve:
Python 3.7 support will soon be dropped from zivid-python, so the README in this repo must be updated. It's easier just to link to the zivid-python repo for supported versions, assuming these samples are supported on all versions supported by zivid-python.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ from the camera can be used.
 
 Note:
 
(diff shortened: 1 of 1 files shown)


johningve:
Decided to keep Python 3.7 support for now, so this PR is obsolete.

PR #224, opened 2026-03-16
Samples: refactor Hand Eye GUI

csu-bot-zivid:
This is a major refactoring of the Hand Eye GUI.
- Improve loading and saving data. It now happens in parallel in background processes.
- Logs the choice of Hand Eye Configuration with the data, so you don't have to infer, remember, or guess.
- A loaded session cannot be modified anymore (can't add data to it).
- Reserve some more space for the residuals so they're not hidden (on many screens).
- Standardize the settings choice flow for Hand-Eye.


TTPinto (approved):
Looks good to me.

